PR #372, opened 2025-09-19
refactor: add generic table prop getters

rubinaga:
## Items
- add table types file to expose the column properties we use in typescript intellisense
- add custom `meta` tag to include properties such as `ariaLabel` and `isExpandable` for columns
- add global table prop helpers
- implement new table prop helpers in profiles (plan to use in all tables eventually)

The aria labels should help with better accessibility and easier locators when testing


yurii-vasyliev (approved):
LGTM

PR #401, opened 2025-10-24
feat: allow wsl profiles to use only_landscape_created

ethanashaw:
This includes a new option in the form to add WSL profiles. It's also present in the edit form, but its disabled, since it can't be edited.

### Manual testing

1. Create a new WSL profile with "Compliance settings" set to "Uninstall..."
2. View the profiles details and see that "Compliance settings" is labeled as "Uninstall..."
3. Edit the profile and see that the "Compliance settings" are disabled and set to "Uninstall..."


rubinaga:
image
@ethanashaw The text is quite long in the options, and the only difference is "Ignore" and "Uninstall". Maybe we could add a helper text for the dropdown:
"This setting applies to WSL child instances that have not been created by Landscape".

Another option would be to have a better label but then the label might get too long. I'm leaning more towards the helper text. What do you think?

PR #404, opened 2025-10-28
feat: add associated instance links for upgrade and removal profiles [LNDENG-2538, LNDENG-2535]

ethanashaw:
The tables for upgrade profiles and removal profiles include the associated instance count, which links to the instances table.


yurii-vasyliev (approved):
LGTM

PR #412, opened 2025-11-10
refactor: enable noUncheckedIndexedAccess

ethanashaw:
I enabled `noUncheckedIndexedAccess`, which adds `undefined` to the type of any item accessed by index. This primarily affects array access (`array[0]`). 

### Avoiding false positives
After enabling the rule, there were many cases causing errors, since the accessed item was not validated. Many of the errors were "false positives" which could be avoided at compile time:
- For mock data, use `const items = [...] as const satisfies T[]` instead of `const items: T[] = [...]`. This results in a tuple type instead of an array type, so items can be accessed safely by index. If `as const` narrows the type too much, you can also add `satisfies T as T` to each item in the array.

I've also includes some narrowing helpers to solve common patterns:
- `pluralizeArray` is similar to `pluralize`, but it takes an array instead of a number. For the singular form, you must pass a function instead of a string.
- `map` allows a tuple to be mapped while preserving its length, since `Array.prototype.length` doesn't.
- `hasProperty` allows you to check if a property is defined in a `Record`-style object. If it is, you'll be able to use that property.

### Potential bugs that were fixed
Other errors revealed potential bugs:
- For testing, there are many situations when you must use `assert` to ensure an item is defined before attempting to use it.
- Combined with style declarations, the errors revealed some references to nonexistent styles.

### Other potential bugs to be aware of
There are some other cases where potential bugs were revealed, but in the interest of avoiding major refactors in this pull request, I didn't fix all of them. Instead, I've used non-null assertions (`!`), which we should not continue to use. I'm also creating Jira issues for them so we don't lose track. Here are the unfixed problems:
- There are some tables with expanded cells, where the table ref is accessed by row index: `EmployeeList`, `InstancesPanel`, `UsnList`, `WslInstancesHeader`, and `RoleList`. (https://warthogs.atlassian.net/browse/LNDENG-3497)
- There are some forms that are used for both a singular item and multiple items. These forms should be split off to improve type safety: `InstalledPackagesActionForm`, `EditSnap`, and `UsnPackagesContainer`. (https://warthogs.atlassian.net/browse/LNDENG-3498)
- The security profile side panel has a step counter. It should be updated to be more type safe: `SecurityProfileAddSidePanel`. (https://warthogs.atlassian.net/browse/LNDENG-3499)

### DIFF
--- a/src/features/upgrades/components/tabPanels/instances/AffectedPackages/AffectedPackages.test.tsx
+++ b/src/features/upgrades/components/tabPanels/instances/AffectedPackages/AffectedPackages.test.tsx
@@ -8,7 +8,7 @@ import { describe, vi } from "vitest";
 import AffectedPackages from "./AffectedPackages";
 
 const increasedLimit = 10;
-const instanceIndex = 13;
+const instanceIndex = 14;
 const instance = instances[instanceIndex];
 const limit = 5;
 const excludedPackages: InstancePackagesToExclude[] = [
--- a/src/tests/server/handlers/scriptProfiles.ts
+++ b/src/tests/server/handlers/scriptProfiles.ts
@@ -3,6 +3,7 @@ import type { Activity } from "@/features/activities";
 import { getEndpointStatus } from "@/tests/controllers/controller";
 import { scriptProfiles } from "@/tests/mocks/scriptProfiles";
 import { http, HttpResponse } from "msw";
+import { ENDPOINT_STATUS_API_ERROR } from "./_constants";
 import { generatePaginatedResponse } from "./_helpers";
 
 export default [
@@ -61,7 +62,7 @@ export default [
           endpointStatus.path.includes("script-profiles/:profileId"))
       ) {
         if (endpointStatus.status === "error") {
-          throw new HttpResponse(null, { status: 500 });
+          throw ENDPOINT_STATUS_API_ERROR;
         }
 
         if (endpointStatus.status === "empty") {
--- a/src/features/instances/constants.ts
+++ b/src/features/instances/constants.ts
@@ -93,7 +93,7 @@ export const STATUS_FILTERS = {
     alternateLabel: undefined,
     filterValue: "unapproved-activities",
     query: "alert:unapproved-activities",
-    icon: { gray: "status-queued", color: undefined },
+    icon: { gray: "status-queued", color: "package-profiles-alert" },
   },
   UpToDate: {
     alertType: "UpToDate",
@@ -109,7 +109,7 @@ export const STATUS_FILTERS = {
     alternateLabel: "WSL profile alert",
     filterValue: "child-instance-profiles",
     query: "alert:child-instance-profiles",
-    icon: { gray: "machines", color: undefined },
+    icon: { gray: "machines", color: "package-profiles-alert" },
   },
 } satisfies Record<string, Status>;
 
--- a/src/features/instances/types/Status.d.ts
+++ b/src/features/instances/types/Status.d.ts
@@ -1,7 +1,7 @@
-type Icon =
-  | { gray: string; color: string }
-  | { gray: string; color: undefined }
-  | { gray: undefined; color: string };
+interface Icon {
+  gray: string | undefined;
+  color: string;
+}
 
 export interface Status {
   icon: Icon;


src/tests/mocks/upgrade-profiles.ts
```diff
@@ -19,7 +19,7 @@ export const upgradeProfiles: UpgradeProfile[] = [
     computers: {
       num_associated_computers: 1,
     },
-  },
+  } as UpgradeProfile,
```
rubinaga: should this also have `satisfies UpgradeProfile as UpgradeProfile`

src/utils/_helpers.ts
```diff
@@ -82,6 +86,14 @@ export const pluralize = (
   return count === 1 ? singularForm : (pluralForm ?? `${singularForm}s`);
 };
 
+export const pluralizeArray = <T>(
```
rubinaga: I dont really see the real benefit of using this over the plain `pluralize`. Do you think it introduces better functionality?

src/features/saved-searches/components/SavedSearchForm/helpers/searchQueryValidation.ts
```diff
@@ -241,7 +238,7 @@ export const validateSearchQuery = (
       continue;
     }
 
-    const token = tokens[i];
+    const token = tokens[i]!;
```
rubinaga: Can't we use `for…of` or a guard like:
```ts
const token = tokens[i];
if (!token) {
  continue;
}
```

Since we've already introduced the non-null assertions rule, I'd prefer if we don't keep doing it

src/pages/dashboard/settings/roles/PermissionBlock/PermissionBlock.tsx
```diff
@@ -39,8 +39,8 @@ const PermissionBlock: FC<PermissionBlockProps> = ({
   const isOptionDisabled = (option: PermissionOption) => {
     return (
       option.values.view === "" ||
-      permissions.includes(
-        options.filter(({ label }) => label === option.label)[0]?.values.manage,
+      (permissions as unknown[]).includes(
```
rubinaga: I think this can be done without type casting to unknown, maybe like this:
```ts
 const isOptionDisabled = (option: PermissionOption) => {
    const manage = options.find(({ label }) => label === option.label)?.values
      .manage;

    if (!manage) {
      return true;
    }

    return option.values.view === "" || permissions.includes(manage);
  };
```

src/features/saved-searches/components/SavedSearchForm/helpers/searchQueryValidation.ts
ethanashaw: I was able to use `for..of` by making a few other changes, let me know what you think (after I push)

src/tests/server/handlers/scriptProfiles.ts
```diff
@@ -50,35 +50,32 @@ export default [
...
+    ({ params }) => {
+      const endpointStatus = getEndpointStatus();
+
+      if (
+        !endpointStatus.path ||
+        (endpointStatus.path &&
+          endpointStatus.path.includes("script-profiles/:profileId"))
+      ) {
+        if (endpointStatus.status === "error") {
+          throw new HttpResponse(null, { status: 500 });
```
rubinaga: I guess this could throw the new `ENDPOINT_STATUS_API_ERROR`

rubinaga (approved):
LGTM but I left a couple comments, if you could have a look

src/features/upgrades/components/tabPanels/instances/AffectedPackages/AffectedPackages.test.tsx
```diff
@@ -1,26 +1,29 @@
...
+const instance = instances[instanceIndex];
 const limit = 5;
 const excludedPackages: InstancePackagesToExclude[] = [
   { id: instance.id, exclude_packages: [] },
 ];
 const onExcludedPackagesChange = vi.fn();
 const onLimitChange = vi.fn();
 const packages = getInstancePackages(instance.id);
+const increasedPackage = packages[increasedLimit - 1];
+assert(increasedPackage);
```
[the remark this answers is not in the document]
ethanashaw: I guess the instance didn't have enough packages, I fixed it now though. For some reason the Vitest extension for VSCode doesn't show errors outside of a `describe`, so I ended up missing it.

PR #416, opened 2025-11-14
feat: add warnings for wsl profiles

ethanashaw:
- Warning notification in WSL profiles page header
- Warning modal when adding WSL profile, only appears once


rubinaga (approved):
LGTM

PR #417, opened 2025-11-14
feat: rename instance statuses

ethanashaw:
- Renamed instance statuses in the status filter dropdown
- Use our own status labels instead of backend alert summaries


rubinaga (approved):
LGTM

PR #432, opened 2025-12-03
fix: Instance page filters changes

rubinaga:
## Changes
- Moved filter button next to searchbox
- Added filter icon to the filter button
- Added background color change when an item is active

## Screenshots
image


ethanashaw (approved):
While you're here, it might be good to add `white-space: nowrap` to the button to fix mobile layouts:

image

Everything else LGTM

PR #435, opened 2025-12-11
Playwright

rubinaga:


e2e/support/fixtures/auth.ts
ethanashaw: Seems like these ESLint errors are false alarms, maybe we need to limit the scope of the plugin to `/src`?

ethanashaw (commented):
The changes look good, but the tests don't seem to pass on my end. I'm not sure if I'm doing something wrong. Do you think we should adjust our scripts in `package.json`?

yurii-vasyliev (approved):
LGTM, all tests passed locally.

PR #436, opened 2025-12-11
Package Management

ethanashaw:
Major changes
- Refactor package installation form and related components to follow new design

Minor changes
- Add `VITE_SELECT_ALL_INSTANCES_ENABLED` environment variable. It controls the "Select all instances on all pages" feature. Right now, it will be disabled by default because it's unsupported by some instance actions on the backend.
- Convert `boldSubstring` helper to a component, `BoldSubstring`, to avoid duplication across helpers
- Add `ref` prop to `LoadingState` to allow it to be used as an intersection observer target in infinite scrolling scenarios
- Replace "Upgrade instances" button with new "Manage packages" actions in the instances page
- Update mock packages handler to enable exact search
- Add `pluralizeWithCount` helper for the common use case of having a number immediately followed by a pluralized word


rubinaga:
@ethanashaw could you take a look at copilot's comments (dismiss or fix in case something makes sense). Also Figma shows the buttons a bit different, parent button is "Manage packages" vs "Deb management", and the items are different too.
image

src/features/packages/components/PackageDropdownSearch/components/PackageDropdownSearchItem/PackageDropdownSearchItem.tsx
```diff
@@ -0,0 +1,44 @@
...
+      <div>
+        <div className="font-monospace">
+          {selectedPackage.name} {selectedPackage.version}
+        </div>
+        <div className="u-text--muted p-text--small u-no-margin">
+          Available on{" "}
+          {pluralize(selectedPackage.computers.count, ["instance"], "exact")}
+        </div>
+      </div>
+      <Button
```
rubinaga, line 31: Can the `Delete` button have a tooltip and hover effect on it
Image

src/features/packages/components/PackageDropdownSearch/PackageDropdownSearch.tsx
rubinaga, line 1: Searching and doing a "space" after removes the bold styling. E.g `apt `. Also when searching `apt 2` (non-exact search) it shows no options in the dropdown
Image

src/features/packages/components/PackageDropdownSearch/components/PackageDropdownSearchList/PackageDropdownSearchList.tsx
```diff
@@ -0,0 +1,113 @@
...
+  if (results.length) {
+    return (
+      <>
+        <ul
+          className={classNames(
+            "p-list u-no-margin p-autocomplete__suggestions",
+          )}
+        >
+          {results.map((item: Package, index: number) => {
+            const disabled = selectedPackages.some(({ id }) => item.id === id);
```
rubinaga, line 62: Not a big deal, more curious why we are "disabling and showing" instead of "not showing" selected items at all in the list 
Image

src/features/packages/components/PackagesActionSummary/PackagesActionSummary.tsx
```diff
@@ -0,0 +1,172 @@
...
+                <div>
+                  <strong className="font-monospace">{packageName}</strong>
+                </div>
+                {packageVersions.map((version) => {
+                  return (
+                    <div key={version.package_id} className={classes.row}>
+                      <span className="font-monospace">
+                        {version.package_version}
+                      </span>{" "}
+                      will be {mapActionToPast(action)} on{" "}
```
rubinaga, line 134: I think we don't need a link for 0 instances affected right?

Image

src/features/packages/components/PackagesActionSummary/components/PackagesActionSummaryDetails/PackagesActionSummaryDetails.tsx
```diff
@@ -0,0 +1,112 @@
...
+
+  const {
+    data: itemsResponse,
+    error: itemsError,
+    isPending: isGettingItems,
+  } = useListPackageChangePlanItems({
+    id: packageChangePlanId,
+    package_ids: [packageChangePlanSummaryItem.package_id],
+    computer_instance_name: search || undefined,
+    limit: DEFAULT_MODAL_PAGE_SIZE,
```
rubinaga, line 45: I know for modals we do 10 entries since we have limited space, but we have more space we could use in the sidepanel. What do you think?

Image

src/features/upgrades/components/UpgradesSummary/UpgradesSummary.tsx
```diff
@@ -0,0 +1,144 @@
...
+import classes from "./UpgradesSummary.module.scss";
+import useNotify from "@/hooks/useNotify";
+import { useOpenActivityDetailsPanel } from "@/features/activities";
+
+interface UpgradesSummaryProps {
+  readonly packageChangePlanId: number;
+  readonly onBackButtonPress?: () => void;
+}
+
+const UpgradesSummary: FC<UpgradesSummaryProps> = ({
```
rubinaga, line 30: Will this table not have the "Upgrade type" column like in the designs? Also when clicking the instances link in the table it opens a sidepanel that's not the same size as the current one which is a bit weird as you can click "back" in the previous sidepanel. I'm assuming it should be the same size (medium).

Image

rubinaga (changes requested):

src/features/packages/components/PackageDropdownSearch/components/PackageDropdownSearchItem/PackageDropdownSearchItem.tsx
gesquivelgaghi: Wait why does it need a tooltip? What would it even say?

rubinaga: "delete" or "remove"

PR #438, opened 2025-12-12
fix: prevent submitting login form twice if error occurs

yurii-vasyliev:
1. Refactor `main.tsx` and `App.tsx` while keeping the same logic
2. Fix submitting login form twice if an error occurs (like incorrect password)
3. Get rid of explicit `authLoading` loading state in favour of react query loading states
4. Split `useUnassigned` into atomic "api" hooks
5. Add/fix some tests based on the changes

This PR affects a lot of important features related to the login flow, so please test everything carefully. Of course, we won't be able to test some login methods on our locals, but still.

### DIFF
--- a/src/features/auth/components/LoginForm/LoginForm.tsx
+++ b/src/features/auth/components/LoginForm/LoginForm.tsx
@@ -53,11 +53,11 @@ const LoginForm: FC<LoginFormProps> = ({ isIdentityAvailable }) => {
             .test({
               message: "Please provide a valid email address",
               test: (value) => {
-                if (value?.match(/\.\./)) {
+                if (!value) {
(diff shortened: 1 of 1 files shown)


ethanashaw (changes requested):
Testing went well, I just have some suggestions.

src/tests/render.tsx
```diff
@@ -5,8 +5,8 @@ import { MemoryRouter, Route, Routes } from "react-router";
 import NotifyProvider, { NotifyContext } from "@/context/notify";
```
rubinaga: Should this file also use `<AppProviders>` to keep the behavior between tests and production aligned?

PR #448, opened 2026-01-15
fix: secondary navigation styling on medium screens

rubinaga:
This PR is related to [this ticket](https://warthogs.atlassian.net/browse/LNDENG-3147)

Before:
image

After:
image


gesquivelgaghi:
out of curiosity, why not move the account settings to the left so there's no gap between the collapsed header and the secondary navigation? it would avoid this asymmetry 
image

PR #460, opened 2026-02-04
fix: instance annotations [LNDENG-3896]

ethanashaw:
I added the missing logic for instance annotations, and also removed some unused helpers and added mock data.


yurii-vasyliev (approved):
LGTM. Thanks for picking it up.

PR #476, opened 2026-02-18
refactor: add new helper functions and replace uses

gesquivelgaghi:
## Summary

- Extracted `pluralizeWithCount` from the Package Management PR, and replaced the fitting `pluralize` uses with it
- Replaced fitting `pluralize` uses with existing `pluralizeArray` function
- Extracted `capitalize` and `formatCountableNoun` from a local helper into the global `_helper` file and replaced fitting uses
- Updated existing tests and added new ones for the new functions

## Release Impact

- **Target Branch**: `dev`
- **Version Impact**:
  - [ ] Patch (Fix)
  - [ ] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist
- [ ] **Changeset Added**: I have run `pnpm changeset` and committed the resulting `.md` file.

### DIFF
--- a/src/pages/dashboard/instances/ReportView/ReportView.tsx
+++ b/src/pages/dashboard/instances/ReportView/ReportView.tsx
@@ -108,7 +108,7 @@ const ReportView: FC<ReportViewProps> = ({ instanceIds }) => {
             {!getInstancesNotUpgradedLoading && (
               <ReportWidget
                 currentCount={upgradedInstanceCount}
-                negativeDescription={`${pluralizeWithCount(Number(getInstancesNotUpgradedResult?.data.length), "instance is", "instances are")} not covered by upgrade profiles.`}
+                negativeDescription={`${pluralizeWithCount(Number(getInstancesNotUpgradedResult?.data.length ?? 0), "instance is", "instances are")} not covered by upgrade profiles.`}
                 positiveDescription={`${pluralizeWithCount(upgradedInstanceCount, "instance is", "instances are")} covered by upgrade profiles.`}
                 title="Upgrade profiles"
                 totalCount={instanceIds.length}
@@ -119,7 +119,7 @@ const ReportView: FC<ReportViewProps> = ({ instanceIds }) => {
             {!getNotPingingInstancesLoading && (
               <ReportWidget
                 currentCount={pingingInstancesCount}
-                negativeDescription={`${pluralizeWithCount(Number(getNotPingingInstancesResult?.data.length), "instance has", "instances have")} not contacted the server within the last 5 minutes.`}
+                negativeDescription={`${pluralizeWithCount(Number(getNotPingingInstancesResult?.data.length ?? 0), "instance has", "instances have")} not contacted the server within the last 5 minutes.`}
                 positiveDescription={`${pluralizeWithCount(pingingInstancesCount, "instance has", "instances have")} contacted the server within the last 5 minutes.`}
                 title="Contacted"
                 totalCount={instanceIds.length}
--- a/src/utils/_helpers.ts
+++ b/src/utils/_helpers.ts
@@ -81,9 +81,9 @@ export const hasOneItem = <T>(array: readonly T[]): array is readonly [T] => {
 export const pluralize = (
   count: number,
   singularForm: string,
-  pluralForm = `${singularForm}s`,
+  pluralForm?: string,
 ) => {
-  return count === 1 ? singularForm : pluralForm;
+  return count === 1 ? singularForm : (pluralForm ?? `${singularForm}s`);
 };
 
 export const pluralizeWithCount = (


src/features/instances/components/TagsAddConfirmationModal/TagsAddConfirmationModal.test.tsx
```diff
@@ -25,7 +25,7 @@ describe("TagsAddConfirmationModal", () => {
         <TagsAddConfirmationModal {...props} tags={testTags} />,
       );
 
-      const title = `Add ${pluralize(testTags.length, `"${testTags[0]}" tag`, `${testTags.length} tags`)} to ${pluralize(props.instances.length, `"${props.instances[0].title}"`, `${props.instances.length} instances`)}`;
+      const title = `Add ${pluralizeArray(testTags, (tag) => `"${tag}" tag`, `tags`)} to ${pluralizeArray(props.instances, (instance) => `"${instance.title}"`, `instances`)}`;
```
ethanashaw: Should we flatten this string and include some assertions for how many items we expect, like the other tests?

src/features/mirrors/components/SeriesPocketList/SeriesPocketList.tsx
```diff
@@ -62,10 +62,7 @@ const SeriesPocketList: FC<SeriesPocketListProps> = ({
         accessor: "package_count",
         Header: "Content",
         Cell: ({ row: { original } }: CellProps<CommonPocket>): ReactNode => (
-          <>{`${original.package_count} ${pluralize(
-            original.package_count,
-            "package",
-          )}`}</>
+          <>{`${pluralizeWithCount(original.package_count, "package")}`}</>
```
ethanashaw:
```suggestion
          <>{pluralizeWithCount(original.package_count, "package")}</>
```

src/features/overview/components/InfoTablesContainer/InfoTablesContainer.tsx
```diff
@@ -227,7 +227,7 @@ const InfoTablesContainer: FC = () => {
                 (row.original.upgrades?.security ?? 0) +
                 (row.original.upgrades?.regular ?? 0);
 
-              return `${packageCount} ${pluralize(packageCount, "package")}`;
+              return `${pluralizeWithCount(packageCount, "package")}`;
```
ethanashaw:
```suggestion
              return pluralizeWithCount(packageCount, "package");
```

src/features/package-profiles/components/PackageProfileDetailsSidePanel/PackageProfileDetailsSidePanel.tsx
```diff
@@ -101,15 +101,15 @@ const PackageProfileDetailsSidePanel: FC = () => {
 
           <InfoGrid.Item
             label="Pending on"
-            value={`${profile.computers.pending?.length ?? 0} ${pluralize(
+            value={`${pluralizeWithCount(
               profile.computers.pending.length,
               "instance",
             )}`}
```
ethanashaw:
```suggestion
            value={pluralizeWithCount(
             profile.computers.pending.length,
             "instance",
           )}
```

src/features/package-profiles/components/PackageProfileList/PackageProfileList.tsx
```diff
@@ -157,7 +157,7 @@ const PackageProfileList: FC<PackageProfileListProps> = ({
         ),
         Cell: ({ row: { original } }: CellProps<PackageProfile>) => (
           <>
-            {`${original.computers["pending"]?.length ?? 0} ${pluralize(original.computers["pending"].length, "instance")}`}
+            {`${pluralizeWithCount(original.computers["pending"].length, "instance")}`}
```
ethanashaw:
```suggestion
            {pluralizeWithCount(original.computers["pending"].length, "instance")}
```

src/features/alert-notifications/components/AlertNotificationsList/AlertNotificationsList.test.tsx
```diff
@@ -65,11 +64,7 @@ describe("AlertNotificationsList", () => {
...
     const button = screen.getByRole("button", {
-      name: `${pendingInstances.length} pending ${pluralize(
-        pendingInstances.length,
-        "computer needs",
-        "computers need",
-      )} authorization`,
+      name: `${pendingInstances.length} pending computers need authorization`,
     });
     expect(button).toBeInTheDocument();
   });
```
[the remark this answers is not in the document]
gesquivelgaghi: Not sure if there's a benefit to the assertion, considering the test isn't actually _about_ testing the scenario where there's multiple pending instances, it more just happens to be the scenario we fall into because the mock we use happens to have multiple instances. But if someone were to change the mock to only be 1 instance, this test shouldn't fail because of an assertion, it should just be changed in order to reflect the updated mock. Does that make sense?

I can also parametize the test to cover both scenarios instead, like the other test discussed below.

src/features/instances/components/TagsAddConfirmationModal/TagsAddConfirmationModal.test.tsx
ethanashaw: I agree that we shouldn't need to worry about testing the phrasing in detail, it's better for us to focus on logic with these tests. Since we've extracted the logic to a helper, I think the tests should ideally pass whether there's one item or multiple.

To be honest, I actually think the test in question is mostly unecessary, because it just checks for declarative content. The only value it provides from a logic or code coverage perspective is that it executes the `getSingularForm` functions we pass to `pluralizeArray`. So I guess there's no need to worry about the tests too much in this case.

src/features/alert-notifications/components/AlertNotificationsList/AlertNotificationsList.test.tsx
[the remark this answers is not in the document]
ethanashaw: Yeah, that is fair, I don't think we should have to worry about the phrasing too much.

PR #508, opened 2026-03-16
Remove Profiles dead code and Repo profiles Applied column

gesquivelgaghi:
## Summary

- Deleted unused components of profile-specific types
- Deleted Applied column in Repository profiles
- Added and updated tests

## Release Impact

- **Target Branch**: `main` (Beta)
- **Version Impact**:
  - [ ] Patch (Fix)
  - [x] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist


ethanashaw (changes requested):
A few QA things before code review

---

When you open a deeper form like "Edit" from the details view, the back button is missing
Image

---

The APT source list for repository profiles has a scrollbar when it's empty
Image

---

The package profile buttons don't seem to fit, this one could be larger I think
image

ethanashaw (changes requested):
One more thing @gesquivelgaghi, the pass rate bar for security profiles gets cut off on smaller screens:

Image

rubinaga:
@gesquivelgaghi 
Some things from me:
1. Repository profiles and Upgrade profiles are missing the columns compliant and non-compliant, was this missed or changed from the Figma file?

2. Should Scripts profiles still be under "Scripts" as a tab or should they have been moved to "Profiles" as a separate page?

3. Navigating back and forth between the profiles, after clicking on "Security profiles" that have limitations, I get a similar notification that no more profiles can be added in all other profiles.
image

4. I don't think we should show "Info" tab when there is only 1 tab in some profiles:
image

5. Small thing, but in the designs the buttons across all profiles are "Add profile" instead of "Add <profile_type> profile", and they have the "+" icon.

6. Not sure how many pockets can be added here, but would it be better to use a `TruncatedCell` instead of extending the row vertically?
image


7. Small thing here too, Upgrade type -> "All upgrades" instead of "All"
image

8. For modals rendered inside sidepanels, I think we need to add the prop `renderInPortal` to have the sidebar show in the background of the modal's backdrop
image

9. Also for description column, maybe a truncated cell is better:
image

10. Can we have the contextual menu open the other way, so it stays inside the sidepanel here, and also add icons, color, separation on the actions? Same for security profiles
image

11. Also since "Edit package constraints" opens a wider panel. I think we want to keep the sidepanel the same size all the time, so the general "View details" would also be "medium" size.

12. When doing -> View removal profile -> Edit removal profile -> add tags -> confirm -> view detail panel is opened, the profile does not show the updated tags immediately. I need to click it from the table again to update it. This is the case in other profiles too.

gesquivelgaghi:
> @gesquivelgaghi Some things from me:
> 
> 1. Repository profiles and Upgrade profiles are missing the columns compliant and non-compliant, was this missed or changed from the Figma file?
 
Changed, the backend doesnt give us this information for these types of profiles.

> 2. Should Scripts profiles still be under "Scripts" as a tab or should they have been moved to "Profiles" as a separate page?

Discussed with Dani, he said to keep them under scripts.

> 3. Navigating back and forth between the profiles, after clicking on "Security profiles" that have limitations, I get a similar notification that no more profiles can be added in all other profiles.

Will look into it, thanks!
 
> 4. I don't think we should show "Info" tab when there is only 1 tab in some profiles:

Should be an easy fix as well.
 
> 5. Small thing, but in the designs the buttons across all profiles are "Add profile" instead of "Add <profile_type> profile", and they have the "+" icon.

Oh, I didn't notice those designs had also been updated. Thanks!

> 6. Not sure how many pockets can be added here, but would it be better to use a `TruncatedCell` instead of extending the row vertically?
> 9. Also for description column, maybe a truncated cell is better:

Will do.
 
> 7. Small thing here too, Upgrade type -> "All upgrades" instead of "All"

Good catch, thanks!

> 8. For modals rendered inside sidepanels, I think we need to add the prop `renderInPortal` to have the sidebar show in the background of the modal's backdrop

Will do

> 10. Can we have the contextual menu open the other way, so it stays inside the sidepanel here, and also add icons, color, separation on the actions? Same for security profiles

Will fix the way it opens, but for the rest, the issue is with the `ResponsiveButtons` component itself, so I think it needs a separate fix. You can see that because when the buttons aren't collapsed, the icons and colors are there.

> 11. Also since "Edit package constraints" opens a wider panel. I think we want to keep the sidepanel the same size all the time, so the general "View details" would also be "medium" size.

Will discuss with Dani, but I think the medium size looks weird with the View profile layout. Also, the Edit and Duplicate side panels are small width too, so they'd have to be changed too, if all are supposed to be the same size. Though to be honest I think it'd look better to make them all small width instead, if it's possible. I think it'll have to be a separate fix though.

> 12. When doing -> View removal profile -> Edit removal profile -> add tags -> confirm -> view detail panel is opened, the profile does not show the updated tags immediately. I need to click it from the table again to update it. This is the case in other profiles too.

Known issue with side panels that I'm already working on.

src/features/profiles/hooks/useCloseManageProfileSidePanel.tsx
```diff
@@ -0,0 +1,12 @@
+import usePageParams from "@/hooks/usePageParams";
+import useSidePanel from "@/hooks/useSidePanel";
+
+export const useCloseManageProfileSidePanel = () => {
+  const { createPageParamsSetter } = usePageParams();
+  const { closeSidePanel } = useSidePanel();
+
+  return () => {
+    createPageParamsSetter({ sidePath: [], profile: "" });
```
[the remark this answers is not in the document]
gesquivelgaghi: deleting this file, its unused

src/features/repository-profiles/helpers.ts
```diff
@@ -0,0 +1,34 @@
...
+      series: [seriesWithPocket],
+    };
+
+    if (existingSeries) {
+      existingSeries.pockets.push(pocketWithDistributionAndSeries);
+    } else if (existingDistribution) {
+      existingDistribution.series.push(seriesWithPocket);
+    } else {
+      distributions.push(distributionWithSeries);
+    }
```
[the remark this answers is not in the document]
gesquivelgaghi: pockets don't exist anymore with deb archive changes so this will be deleted.

src/features/profiles/api/useRemoveProfile.ts
```diff
@@ -0,0 +1,84 @@
...
+      case ProfileTypes.removal:
+        return removeRemovalProfileQuery.isPending;
+      case ProfileTypes.script:
+        return isArchivingScriptProfile;
+      case ProfileTypes.security:
+        return isArchivingSecurityProfile;
+      case ProfileTypes.upgrade:
+        return removeUpgradeProfileQuery.isPending;
+      case ProfileTypes.wsl:
+        return isRemovingWslProfile;
```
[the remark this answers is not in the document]
gesquivelgaghi: added

src/features/profiles/components/ViewProfileSidePanel/components/ViewProfileGeneralBlock/ViewProfileGeneralBlock.tsx
```diff
@@ -0,0 +1,60 @@
...
+  if (isGettingAccessGroups) {
+    return <SidePanel.LoadingState />;
+  }
+
+  const status = isProfileArchived(profile) ? "Archived" : "Active";
+
+  return (
+    <Blocks.Item title="General">
+      <InfoGrid dense>
+        <InfoGrid.Item label="Name" value={profile.title} />
```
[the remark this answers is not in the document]
gesquivelgaghi: nope, we want to display the title but the UI will refer to it as name

src/features/repository-profiles/components/ViewRepositoryProfileAptSourcesTab/ViewRepositoryProfileAptSourcesTab.tsx
```diff
@@ -0,0 +1,56 @@
...
+}
+
+const ViewRepositoryProfileAptSourcesTab: FC<
+  RepositoryProfileDetailsAptSourcesTabProps
+> = ({ profile }) => {
+  const { aptSources, isGettingAPTSources } = useGetAPTSources();
+  const profileSources = aptSources.filter((source) =>
+    source.profiles.includes(profile.name),
+  );
+
```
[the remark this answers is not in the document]
gesquivelgaghi: aptSources does default to an empty array inside `useGetAPTSources`

src/features/repository-profiles/components/ViewRepositoryProfileAptSourcesTab/ViewRepositoryProfileAptSourcesTab.tsx
```diff
@@ -0,0 +1,56 @@
...
+        Cell: ({ row: { original: source } }: CellProps<APTSource>) =>
+          source.line,
+      },
+    ],
+    [],
+  );
+
+  if (isGettingAPTSources) {
+    return <LoadingState />;
+  }
```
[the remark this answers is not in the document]
gesquivelgaghi: same as above

src/features/profiles/components/ProfileAssociatedInstancesLink/ProfileAssociatedInstancesLink.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+import { describe, expect, it, vi } from "vitest";
+import ProfileAssociatedInstancesLink from "./ProfileAssociatedInstancesLink";
+import useSidePanel from "@/hooks/useSidePanel";
+import { profiles } from "@/tests/mocks/profiles";
+
+vi.mock("@/hooks/useSidePanel", () => ({
+  default: vi.fn(),
+}));
+
+vi.mock("@/components/layout/NoData", () => ({
```
rubinaga: You dont need to mock this, you can use the `NO_DATA_TEXT` directly

src/features/profiles/components/ProfileAssociatedInstancesLink/ProfileAssociatedInstancesLink.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+  default: () => <span>No data</span>,
+}));
+
+vi.mock("@/features/wsl-profiles", () => ({
+  WslProfileNonCompliantInstancesList: () => <div>WSL list</div>,
```
rubinaga: Removing this did not make a difference when I ran the tests, is there a reason we need it?

src/features/profiles/components/ProfileAssociatedInstancesLink/ProfileAssociatedInstancesLink.test.tsx
```diff
@@ -0,0 +1,188 @@
...
+      <ProfileAssociatedInstancesLink
+        count={3}
+        profile={wslProfile}
+        query="wsl:1:noncompliant"
+      />,
+    );
+
+    await userEvent.click(screen.getByRole("button", { name: "3 instances" }));
+
+    expect(setSidePanelContent).toHaveBeenCalledWith(
```
rubinaga: why not check that the sidepanel has been opened directly by getting its title, instead of mocking the function and checking its parameters

src/features/profiles/components/ProfileAssociatedInstancesLink/ProfileAssociatedInstancesLink.tsx
```diff
@@ -0,0 +1,99 @@
...
+  count,
+  profile,
+  query,
+  isPending = false,
+  isGeneralAssociation = false,
+}) => {
+  const { setSidePanelContent } = useSidePanel();
+
+  if (isPending) {
+    return <Spinner />;
```
rubinaga: You could use `<LoadingState inline />` here

src/features/profiles/components/ProfileAssociatedInstancesLink/ProfileAssociatedInstancesLink.tsx
```diff
@@ -0,0 +1,99 @@
...
+  if (count === 0) {
+    return <>0 instances</>;
+  }
+
+  const text =
+    isGeneralAssociation && profile.all_computers
+      ? "All instances"
+      : pluralizeWithCount(count, "instance");
+
+  const packageComplianceIds = (() => {
```
rubinaga: I'd say we dont usually use IIFE in the codebase and you could extract this to a separate function, but thats minor

src/features/profiles/components/ProfilesContainer/ProfilesContainer.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+vi.mock("@/components/layout/TablePagination", () => ({
+  TablePagination: ({ totalItems }: { readonly totalItems: number }) => (
+    <div>{`pagination:${totalItems}`}</div>
+  ),
+}));
+
+const mockUsePageParams = vi.mocked(usePageParams);
+const mockUseProfiles = vi.mocked(useProfiles);
+
+const props = {
```
rubinaga: could add `const props: ComponentProps<typeof ProfilesContainer>` here and in general in other places

src/features/profiles/components/ProfilesContainer/ProfilesContainer.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+import { screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import ProfilesContainer from "./ProfilesContainer";
+import { ProfileTypes } from "../../helpers";
+import type { Profile } from "../../types";
+import usePageParams from "@/hooks/usePageParams";
+import useProfiles from "@/hooks/useProfiles";
+import { profiles as profileList } from "@/tests/mocks/profiles";
+
+vi.mock("@/hooks/usePageParams", () => ({
```
rubinaga: do we need to do all these hook mocks? why not use msw handlers and mocks

src/features/profiles/components/AddProfileButton/AddProfileButton.tsx
```diff
@@ -1,15 +1,26 @@
...
+const LOCAL_STORAGE_ITEM = "_landscape_isWslPopupClosed";
+
+interface AddProfileButtonProps {
+  readonly isWsl?: boolean;
+  readonly isInsideScriptHeader?: boolean;
 }
 
-const WslProfileAddButton: FC<WslProfileAddButtonProps> = ({ disabled }) => {
+const AddProfileButton: FC<AddProfileButtonProps> = ({
+  isWsl = false,
```
rubinaga: you could probably infer this information from the url instead of passing it as props

src/features/profiles/components/ProfilesList/components/AssociatedInstancesCell/AssociatedInstancesCell.test.tsx
```diff
@@ -0,0 +1,48 @@
+import { renderWithProviders } from "@/tests/render";
+import { screen } from "@testing-library/react";
+import { describe, expect, it, vi } from "vitest";
+import { ProfileTypes } from "../../../../helpers";
+import AssociatedInstancesCell from "./AssociatedInstancesCell";
+import { useGetProfileAssociatedCount } from "../../../../hooks/useGetProfileAssociatedCount";
+import { profiles } from "@/tests/mocks/profiles";
+
+vi.mock("../../../../hooks/useGetProfileAssociatedCount", () => ({
```
rubinaga: again do we need to mock the hooks here

src/features/repository-profiles/components/RepositoryProfileForm/RepositoryProfileForm.tsx
```diff
@@ -17,10 +17,10 @@ import { useAPTSources } from "@/features/apt-sources";
 import useDebug from "@/hooks/useDebug";
```
rubinaga, line 16: could you also add a "back" button on edit in this form (when opened from details)

src/pages/dashboard/profiles/repository-profiles/RepositoryProfilesPage.tsx
```diff
@@ -0,0 +1,96 @@
...
+
+const RepositoryProfilesPage: FC = () => {
+  const { getRepositoryProfilesQuery } = useRepositoryProfiles();
+  const { data: getRepositoryProfilesResult, isPending } =
+    getRepositoryProfilesQuery();
+  const { sidePath, lastSidePathSegment, createPageParamsSetter } =
+    usePageParams();
+  const { setIsProfileLimitReached } = useProfiles();
+
+  useEffect(() => {
```
rubinaga: could there be a way to avoid calling this useEffect on pages that dont have a limit?

rubinaga (changes requested):
@gesquivelgaghi I made some comments here, mainly ones on the tests, I didnt comment on all files but if you could swap the mock usage especially for hooks or components to msw handlers it would be nice

src/pages/dashboard/profiles/repository-profiles/RepositoryProfilesPage.tsx
gesquivelgaghi: It's needed to avoid the `isProfileLimitReached` value from staying true after navigating from a different profile to one where the limit is not reached

rubinaga: I meant more if it could be centralized in the profiles context by resetting it on navigation change or something like that. And then the pages that need to set it can do so, the others dont have to

src/features/profiles/components/ProfilesContainer/ProfilesContainer.test.tsx
gesquivelgaghi: I removed the pageParams mock, but can't see a way to not need the `useProfiles` mock, since its a context hook so msw cant really help.

But I'm curious, why should we avoid mocking non-api hooks and other components? Tbh copilot generated the tests with everything mocked, and I did delete some particularly unnecessary ones before committing (though I clearly missed others, which I appreciate your comments for pointing those out). But I thought since these are unit tests, decoupling them from their dependencies isn't really a bad thing, so I left the ones I thought were useful/made the tests more maintainable.

src/features/package-profiles/api/useGetPackageProfile.ts
```diff
@@ -8,7 +9,15 @@ export interface UseGetPackageProfileParams {
...
-export const useGetPackageProfile = (name: string) => {
+export const useGetPackageProfile = (
+  name: string,
+  config: Omit<
+    UseQueryOptions<
+      AxiosResponse<{ result: [PackageProfile] }, AxiosError<ApiError>>
+    >,
+    "queryKey" | "queryFn"
+  > = {},
+) => {
```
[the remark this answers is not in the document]
gesquivelgaghi: fixed

src/pages/dashboard/profiles/repository-profiles/RepositoryProfilesPage.tsx
gesquivelgaghi: done

src/features/profiles/components/ProfilesContainer/ProfilesContainer.test.tsx
rubinaga: I think our unit tests arent strictly unit tests, more like integration tests. I believe it helps catch a bigger amount of bugs. And I think wherever we can use the actual components / hooks its better to use them and we have (most of) the tools in place for them. Like for `usePageParams` you can pass url parameters to `renderWithProviders` if you need to consume url parameters in the component. Obviously sometimes its inevitable to mock something but in general I think our hook mocks have been scoped to auth, account, environment stuff.

rubinaga:
@gesquivelgaghi could you take a look at this re-rendering issue too please? I suspect its the Profile context causing the sidebar to re-render. It doesn't happen in `main`.
[Screencast from 2026-04-16 10-19-41.webm](https://github.com/user-attachments/assets/01e8b11a-2270-4c51-8b27-9ec9c1de954a)

src/features/profiles/api/useRemoveProfile.ts
```diff
@@ -0,0 +1,86 @@
...
+    useRemoveRebootProfile();
+  const { removeWslProfile, isRemovingWslProfile } = useRemoveWslProfile();
+
+  const { removePackageProfileQuery } = usePackageProfiles();
+  const { removeRepositoryProfileQuery } = useRepositoryProfiles();
+  const { removeRemovalProfileQuery } = useRemovalProfiles();
+  const { removeUpgradeProfileQuery } = useUpgradeProfiles();
+
+  const removeProfile = async (params: RemoveProfileParams) => {
+    switch (type) {
```
yurii-vasyliev: All hooks are initialized regardless of type. Every call to this hook sets up mutation observers for all profile types, even though only one type is being used. This is unnecessary overhead that grows with each new profile type.

src/features/profiles/api/useRemoveProfile.ts
```diff
@@ -0,0 +1,86 @@
...
+
+interface RemoveWithProfileIdParams {
+  id: number;
+}
+
+interface RemoveWithProfileNameParams {
+  name: string;
+}
+
+export type RemoveProfileParams = RemoveWithProfileNameParams &
```
yurii-vasyliev: I believe, this should be an intersection, not a union. It's either "id" or "name".

src/features/profiles/api/useRemoveProfile.ts
gesquivelgaghi: Yeah, I'm aware of this issue but wasn't really sure how to bypass it. Maybe I can call the individual `useRemove[Type]Profile` hooks in the individual `[Type]ProfilesPage` and then store it in the `ProfilesContext` so it can be accessed inside the `RemoveProfileModal`

src/features/profiles/components/ProfilesContainer/ProfilesContainer.test.tsx
gesquivelgaghi: makes sense, i removed the unnecessary mocks and left ones that significantly simplify the boilerplate needed

src/features/profiles/api/useRemoveProfile.ts
gesquivelgaghi: i tried that first, but it leads to a typescript issue since this is used for the general `removeProfile` mutation that is assigned each individual profile's `mutateAsync` depending on the case. So it needs to include both `name` and `id` as required params or typescript will say that, for example, `removeWslProfile` can't be used as `removeProfile` because `id: number` is required for the first and optional for the latter.

src/features/profiles/api/useRemoveProfile.ts
gesquivelgaghi: that worked. its not super ideal, since now we're initializing the hook for the remove/archiveProfile before its needed, but at least its just for the profile the user's on

yurii-vasyliev: The only way to do is separate hooks into `useRemovePackageProfile`, `useRemoveRepositoryProfile`, etc. This way each hook will already know the type. And put them into each "profile type" folder separately. We can still have a shared "profiles" folder, but for things that are different across profiles we should use subfolders for each profile type individually.

src/features/profiles/api/useRemoveProfile.ts
yurii-vasyliev: I think this issue will be gone too when we split hooks.

src/features/profiles/api/useRemoveProfile.ts
gesquivelgaghi: The hooks already exist separately, the issue is they each need to be used within `RemoveProfileModal` (which is a general component for all profile types), and hooks can't be conditionally called.

What I did was instead call each one individually inside `PackageProfilesPage`, `RepositoryProfilesPage`  etc, and then save the needed values to the ProfilesContext to be used by `RemoveProfileModal` without needing to invoke the hooks themselves there. The issue with this approach is that the `useRemovePackageProfile` hook will be initialized as soon as the user navigates to the package profiles page, regardless of whether or not they intend to remove a profile. But it'll be limited to the type of the page the user is on, which I think is an improvement on the previous solution

src/hooks/useProfiles.ts
yurii-vasyliev, line 1: I'd suggest moving this hooks inside the feature folder.

rubinaga:
these changes have been merged with deb archive right? is there anything extra needed from this PR or should it be closed @gesquivelgaghi

gesquivelgaghi:
> these changes have been merged with deb archive right? is there anything extra needed from this PR or should it be closed @gesquivelgaghi

I updated the PR with the debarchive changes so as it stands now these are the changes that are left that didn't get merged with debarchive. The vast majority (about 125 out of 135 files) are deletions/one-liners, but I'm still integrating them with the repo profiles changes

src/features/profiles/components/ViewProfileSidePanel/components/ViewProfileActionsBlock/ViewProfileActionsBlock.test.tsx
```diff
@@ -0,0 +1,172 @@
+import { renderWithProviders } from "@/tests/render";
+import { setScreenSize } from "@/tests/helpers";
+import { screen } from "@testing-library/react";
+import userEvent from "@testing-library/user-event";
+import { describe, expect, it } from "vitest";
```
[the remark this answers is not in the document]
gesquivelgaghi: we never import `beforeEach`

src/components/layout/ResponsiveTable/ResponsiveTable.tsx
```diff
@@ -8,6 +8,8 @@ import { type JSX } from "react";
 import { useMediaQuery } from "usehooks-ts";
 import classes from "./ResponsiveTable.module.scss";
 
+const DEFAULT_MIN_TABLE_WIDTH = 950;
```
yurii-vasyliev, line 11: We're changing the default min width, while some places use the old default value (1024). Please check all those places.

.changeset/clear-poems-taste.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": minor
+---
+
+Standardized the information displayed across different profiles pages
```
yurii-vasyliev: The removal of the Applied column is a feature removal users may notice, please add it here.

src/components/layout/Chip/Chip.tsx
```diff
@@ -11,7 +11,7 @@ const Chip: FC<ChipProps> = ({ className, value }) => {
   return (
     <span
       className={classNames(
-        "p-chip is-dense u-no-margin--bottom u-no-margin--right",
```
yurii-vasyliev, line 14: This might be dangerous. We need to check every single place where we use the `Chip` component.

src/features/repository-profiles/components/RepositoryProfileList/RepositoryProfileList.tsx
```diff
@@ -104,29 +101,15 @@ const RepositoryProfileList: FC<RepositoryProfileListProps> = ({
...
-        className: classes.pending,
         meta: {
           ariaLabel: ({ original }) =>
             `${original.title} profile pending machines count`,
         },
-        Cell: ({ row: { original } }: CellProps<RepositoryProfile>) => (
-          <>{pluralize(original.pending_count ?? 0, ["instance"], "exact")}</>
-        ),
+        Cell: ({ row: { original } }: CellProps<RepositoryProfile>) =>
+          pluralize(original.pending_count, ["instance"], "exact"),
```
yurii-vasyliev, line 112: We're 100% sure now, that's is gonna be number all the time, right?

src/styles/partials/table.scss
```diff
@@ -14,11 +14,7 @@
...
 }
 
-.checkbox-cell {
-  width: 2rem;
-}
-
-.date-cell {
+.large-cell {
   width: 14rem;
 }
```
ethanashaw, line 19: It's not really in the scope of these changes, but I think it's better to avoid global classes as much as possible, since they are harder to trace. Maybe that's something we can come back to later.

ethanashaw (dismissed):

src/styles/partials/table.scss
gesquivelgaghi: fair enough. in this case I just used copilot to find all the usages, but it can be unreliable i suppose.

src/features/repository-profiles/components/RepositoryProfileList/RepositoryProfileList.tsx
gesquivelgaghi: yes, it's explicitly typed as a non-optional number

src/components/layout/Chip/Chip.tsx
gesquivelgaghi: I did check, it was only used in a few other places. Specifically, `AutoinstallFilesList`, `AutoinstallFileSidePanelTitle` and `InfoPanel`, and with the default `0.5rem` margin, it doesn't break any of these uses.
image
image
image

There's a separate issue where the chip in the `AutoinstallFileSidePanelTitle` isn't `inline`, but that's unrelated. I made a ticket for it https://warthogs.atlassian.net/browse/LNDENG-4603

src/components/layout/ResponsiveTable/ResponsiveTable.tsx
gesquivelgaghi: i changed the ones that needed a higher width. I think long term we need to improve how we set the table/cell width in general, but for now they all look fine in the UI.

ethanashaw (approved):

PR #531, opened 2026-03-27
Refactor instance modals

ethanashaw:
## Summary

The shut down and restart modals would not close after submitting.

Now, you can also specify the delivery time for the instances page actions.

## Release Impact

- **Version Impact**:
  - [ ] Patch (Fix)
  - [x] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist


rubinaga (changes requested):
Can you also add this change in the individual instance view (in the InfoPanel)

yurii-vasyliev (approved):
LGTM. Just one minor thing you overlooked.

rubinaga (commented):
You can remove the "createPortal" and add a `renderInPortal` prop directly to the ConfirmationModal instead. Also will you add the "deliver after" in the bulk actions too?

rubinaga (approved):
@ethanashaw a minor thing, when submitting either modal and reopening it, the previous state remains there instead of resetting. Can you fix that before merging? Otherwise LGTM

ethanashaw:
> @ethanashaw a minor thing, when submitting either modal and reopening it, the previous state remains there instead of resetting. Can you fix that before merging? Otherwise LGTM

Good catch, thanks

PR #545, opened 2026-04-09
Add icons to collapsed buttons inside dropdowns

gesquivelgaghi:
## Summary

- Added icons to the buttons inside the dropdown of `ResponsiveButtons`
- Changed chevron icon to be on the left when `ResponsiveDropdownItem` opens to the left

## Release Impact

- **Target Branch**: `main` (Beta)
- **Version Impact**:
  - [x] Patch (Fix)
  - [ ] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist

## UI Changes

BEFORE:
image
image
image

<br />
<br />

AFTER: 
image
image image

### DIFF
--- a/src/components/ui/ResponsiveButtons/ResponsiveButtons.test.tsx
+++ b/src/components/ui/ResponsiveButtons/ResponsiveButtons.test.tsx
@@ -113,4 +113,18 @@ describe("ResponsiveButtons", () => {
       screen.getByRole("button", { name: "Disabled button" }),
     ).toHaveAttribute("aria-disabled");
   });
+
+  it("preserves icon children in collapsed menu items", async () => {
+    setScreenSize("xs");
+
+    render(<ResponsiveButtons buttons={buttons} />);
+
+    await userEvent.click(screen.getByRole("button", { name: /actions/i }));
+
+    const buttonWithIcon = screen.getByRole("button", {
+      name: "Button with icon",
+    });
+
+    expect(buttonWithIcon.querySelector(".p-icon--code")).toBeInTheDocument();
+  });
 });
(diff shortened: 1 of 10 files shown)


rubinaga (commented):
Image

@gesquivelgaghi I think it would be nice if the chevron was "chevron left" here and in line with the other icons (on the left). What do you think?

gesquivelgaghi:
@rubinaga the reason I left it as is, is because it the dropdown options are about to change to these: 
image
where no options will have icons, and when the buttons are collapsed, they'll all have the chevron at the end of the right side. I don't know about changing the chevron because that's the standard for all our button dropdowns, so if we were to change it we would lose that UI consistency. But the dropdown itself will open on the left side if there's not enough space on the right, like so: 
image

rubinaga:
@gesquivelgaghi won't that somehow be a bit worse in that case when all are combined under "Actions"? because the chevron will be pointing to the right, but there will never be space to be opened to the right so they always will end up left. I think this pattern is only in this "Instances page" in the "actions", and in the "filters" so I don't think we have a set pattern in the project yet. I don't think its a big deal if one points left and the other right but could be worth bringing up with Dani I guess

src/components/ui/ResponsiveButtons/ResponsiveButtons.test.tsx
```diff
@@ -113,4 +113,18 @@ describe("ResponsiveButtons", () => {
...
+    render(<ResponsiveButtons buttons={buttons} />);
+
+    await userEvent.click(screen.getByRole("button", { name: /actions/i }));
+
+    const buttonWithIcon = screen.getByRole("button", {
+      name: "Button with icon",
+    });
+
+    expect(buttonWithIcon.querySelector(".p-icon--code")).toBeInTheDocument();
+  });
```
marqode, line 256: Test coverage shows 67%- it looks like the easiest win would be adding a test for grouped={false} in props (or just ask copilot to increase test coverage).

marqode (changes requested):
Looking good, the only thing I would add is more test coverage of ResponsiveButtons.

marqode (approved):
LGTM!

rubinaga (approved):
LGTM

rubinaga:
@gesquivelgaghi please resolve the testing conflicts before merging

PR #540, opened 2026-04-08
Soft deletion [LNDENG-3984]

ethanashaw:
## Summary

I updated the `useRemoveInstancesFromLandscape` hook to use the new soft deletion endpoint. It's not merged on the dev site, so it can't be tested without MSW yet.

## Release Impact

According to the [Landscape Server Release Cycle](https://docs.google.com/document/d/1sKAp5IvArpfArhMNojFwKOHm9LEdHKB4Et6tu1_-0GY/edit?tab=t.0), this change will target the following release cycle:
- **Target Branch**: `dev` / `main` (Beta)
- **Version Impact**:
  - [ ] Patch (Fix)
  - [x] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist
- [x] **Changeset Added**: I have run `pnpm changeset` and committed the resulting `.md` file.
- [x] **UI Verified**: I have verified the changes locally.
- [x] **Linting**: No linting errors are present (especially in `scripts/`).

PR #547, opened 2026-04-10
Soft deletion [LNDENG-3984]

jmucc:
## Summary

Adds support for the v2 deletion endpoint. I forked off Ethan's [branch](https://github.com/canonical/landscape-ui/pull/540) and changed a couple lines as a result of the manual testing.

## Release Impact

- **Version Impact**:
  - [ ] Patch (Fix)
  - [x] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist


gesquivelgaghi (approved):

jmucc:
Gabe pointed out some unit tests are failing and we might want to add a test for calling the endpoint with more computer_ids than the limit. This shouldn't be merged until those matters are addressed.

jmucc:
Changes since Friday:
- Fixed the unit tests
- Removed check against endpoint limit. I chatted with Rubin and he mentioned this isn't enforced in other places in the application, (e.g. attach-token) because of the UI-imposed limit being less anyways. So leaving it out for consistency.
- Brought back the old endpoint and choose between legacy and REST endpoint depending on the feature flag. This is how it should have been done initially, I only realized this recently.

Things that might be worth a second look:
- I changed the case on the `ubuntu-pro-licensing` feature. To me this seemed inconsistent with the backend response and potentially a bug.
- Examining the browser tools I see that the `features` and `about` endpoints are getting called twice. I think this could be as a result of me including the `useAuth` hook in the `useRemoveInstancesFromLandscape` hook, but I do at least see the `features` endpoint being duplicated in staging.

I initially considered this work pretty time-sensitive in order to start our testing. But on second thought I'd rather make sure we get it right on the first go then add bugs we'll have to fix later. I'm happy to address review comments or leave it to the frontend team from here if there are any doubts. Thanks.

src/features/instances/api/useRemoveInstancesFromLandscape.test.tsx
yurii-vasyliev, line 1: We test hooks in a bit different (simplified) way which can be seen here: `src/hooks/__tests__/useFeatures.test.tsx`, not with direct API hits.
While this is fine, separate API tests is a feature we'll be working on next cycle.

jmucc: Okay, I removed those tests since they were API dependent.

ethanashaw (commented):
@jmucc, do we still need to include a warning for the 1,000 instance limit?

jmucc:
@ethanashaw I took it out based on a conversation I had with Rubin. We could keep it for safety for sure but I don't think there are limits being enforced anywhere in the UI yet.
 

> Hey, we're already limiting the amount of instances that can be selected to be max 100. So /attach-token won't exceed that 500 limit. I think eventually this is a bigger discussion to be had when we allow of selection of more instances (e.g all possible instances). Then we will need to coordinate on limits on the endpoints.

ethanashaw:
> @ethanashaw I took it out based on a conversation I had with Rubin. We could keep it for safety for sure but I don't think there are limits being enforced anywhere in the UI yet.

Sounds good, we can leave it out.

ethanashaw (approved):

src/types/FeatureKey.ts
```diff
@@ -1,9 +1,10 @@
...
+  | "oidc-configuration"
   | "script-profiles"
-  | "usg-profiles"
+  | "spa-dashboard"
   | "support-provider-login"
-  | "wsl-child-instance-profiles"
-  | "ubuntu_pro_licensing";
+  | "ubuntu-pro-licensing"
+  | "usg-profiles"
+  | "wsl-child-instance-profiles";
```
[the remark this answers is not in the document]
ethanashaw: That's fine, it's a very minor change, I think we can keep it in this PR

PR #564, opened 2026-04-21
fix: CSS ordering inconsistency between dev and production build

yurii-vasyliev:
## Problem

In Vite dev mode, CSS from SCSS modules (e.g. `TableFilter.module.scss`) is injected as inline `<style>` tags following the JS module import order. This means custom classes like `.toggle` naturally appear after Vanilla Framework styles and win the cascade.
In production builds, Vite extracts CSS into separate files where chunk ordering is not guaranteed. Vanilla Framework's button selectors (e.g. `.p-button--base with margin: 0 1rem 1.25rem 0`) can appear **after** our module styles, overriding them due to equal specificity + later source order. This causes visual regressions that only appear in deployed environments, but not on localhost.

## Solution

Wrap Vanilla Framework's CSS output in a @layer vanilla block. Per the CSS cascade spec, **unlayered styles always take priority over layered styles**, regardless of specificity or source order. This means that all CSS Modules and custom partials automatically win over Vanilla's rules without needing specificity hacks like doubled selectors (`.toggle.toggle`) or `!important`.

## Release Impact

- **Version Impact**:
  - [x] Patch (Fix)
  - [ ] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist
- [ ] **Linting**: No linting errors are present (especially in `scripts/`).

### DIFF
--- a/.changeset/ten-moles-hang.md
+++ b/.changeset/ten-moles-hang.md
@@ -1,0 +1,5 @@
+---
+"landscape-ui": patch
+---
+
+Fix CSS ordering inconsistency between dev and production builds
--- a/eslint.config.mjs
+++ b/eslint.config.mjs
@@ -12,6 +12,7 @@ export default defineConfig(
     ignores: [
       "dist",
       "dist-ssr",
+      "assets",
       "node_modules",
       "playwright-report",
       "playwright",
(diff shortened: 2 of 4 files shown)


marqode (approved):
LGTM!

PR #575, opened 2026-04-29
refactor api endpoints, fix publication tests, fix publication form cancel button

rubinaga:
## Summary

Fixed publication apis to use `@canonical/landscape-openapi`
Removed duplicate endpoints from publications
Moved batch endpoints to respective features
Added test fixes for publications
Fixed some build errors (there are more from profiles that I have little context on, I did not scope them into this PR)
Fixed cancel button not working in publication form
Fixed `Add publication` button showing in the page header when empty state was shown
Changed publication column from using `label` to `displayName`
Added server side filtering to publications
Changed query structure from Mirrors, Locals, Publication targets that filters Publications 

-- Additional:
Removed types from `/publication-targets` that were re-exports of `@canonical/landscape-openapi`
Changed imports in `/publication-targets` to import directly from `@canonical/landscape-openapi`

## Release Impact

- **Version Impact**:
  - [ ] Patch (Fix)
  - [ ] Minor (Feature)
  - [ ] Major (Breaking)

## Checklist
- [ ] **Changeset Added**: I have run `pnpm changeset` and committed the resulting `.md` file.
- [ ] **UI Verified**: I have verified the changes locally.
- [ ] **Linting**: No linting errors are present (especially in `scripts/`).


src/features/local-repositories/components/PublishLocalRepositorySidePanel/components/PublishRepositoryNewForm/PublishRepositoryNewForm.tsx
```diff
@@ -101,10 +107,12 @@ const PublishRepositoryNewForm: FC<PublishRepositoryNewFormProps> = ({
   const publicationTargetOptions = useMemo<SelectOption[]>(
     () => [
       { label: "Select publication target", value: "" },
-      ...publicationTargets.map((publicationTarget) => ({
-        label: publicationTarget.displayName,
-        value: publicationTarget.name,
-      })),
+      ...publicationTargets
+        .filter((pt) => Boolean(pt.name))
```
gesquivelgaghi: why is this filter needed? in real usage, won't the `name` always be there?

rubinaga: ah it was an intermediate solution when solving `name` being undefined, removed it, thanks for the feedback

src/features/publications/components/PublicationDetails/PublicationDetails.tsx
```diff
@@ -91,7 +92,7 @@ const PublicationDetails = ({
             <InfoGrid.Item
               label="Architectures"
               large
-              value={publication.architectures.join(", ")}
+              value={publication.architectures?.join(", ") || <NoData />}
```
gesquivelgaghi: nitpick: infogrid item already adds the `<NoData />` fallback so this is redundant

src/pages/dashboard/repositories/publications/PublicationsPage.tsx
```diff
@@ -37,6 +42,12 @@ const PublicationsPage: FC = () => {
             <AddPublicationForm />
           </SidePanel.Suspense>
         )}
+        {lastSidePathSegment === "add-target" && (
+          <SidePanel.Suspense key="add-target">
+            <SidePanel.Header>Add publication target</SidePanel.Header>
+            <AddPublicationTargetForm />
+          </SidePanel.Suspense>
+        )}
```
gesquivelgaghi, line 50: Are we really supposed to have the add publication-target form open from the publications page? I thought we were supposed to just navigate to the publication-targets page with `sidePath=add`. I think both mirrors and locals have a similar scenario, so we should make sure the behavior is consistent, whatever we go with. Might be worth asking Dani

src/features/publications/components/PublicationsContainer/PublicationsContainer.tsx
```diff
@@ -67,31 +67,50 @@ const PublicationsContainer = () => {
...
+          </>
+        )}
+      </>
+    );
+  };
+
+  const actions =
+    isGettingPublicationTargets || isMissingTargets
+      ? []
+      : [<AddPublicationButton key="add-publication-button" />];
```
gesquivelgaghi, line 114: the button at the top shouldn't be there in the Publications Empty page either (discussed in https://www.figma.com/design/IgBhJyoWLf4y05pIyFixdk?node-id=319-2841#1709287461 and https://chat.canonical.com/canonical/pl/3u9kg5mad7ds7jz58ktyzcwz9r)

src/features/publications/api/useDeletePublication.ts
```diff
@@ -1,19 +1,23 @@
...
+  DeletePublicationData,
+  DeletePublicationError,
+  DeletePublicationResponse,
+} from "@canonical/landscape-openapi";
 import { useMutation, useQueryClient } from "@tanstack/react-query";
 import type { AxiosError, AxiosResponse } from "axios";
 
 interface DeletePublicationParams {
-  publicationName: string;
+  publicationName: DeletePublicationData["path"]["name_2"];
```
gesquivelgaghi: i assume this `_2` suffix comes from the new changes to the openapi generation? Is the naming stable or will the suffix number change with other updates? if it's prone to change maybe we should just keep it as a string? Not sure what the situation is

src/libs/pageParamsManager/types.ts
```diff
@@ -33,6 +33,7 @@ export interface PageParams {
   code: string;
   sidePath: string[];
   name: string;
+  profile: string;
```
gesquivelgaghi: seems like marc is changing this so all profiles use `name` instead, so maybe we can skip this change and let marc's profiles pr fix this part of the errors

src/libs/pageParamsManager/constants.ts
```diff
@@ -140,6 +140,11 @@ export const PARAMS_CONFIG: ParamsConfig = [
     shouldResetPage: false,
     defaultValue: DEFAULT_EMPTY_STRING,
   },
+  {
+    urlParam: "profile",
+    shouldResetPage: false,
+    defaultValue: DEFAULT_EMPTY_STRING,
+  },
```
gesquivelgaghi: same here, marc's pr should fix the errors caused by this

src/features/publications/components/RepublishPublicationModal/RepublishPublicationModal.tsx
marqode, line 39: Also set sidePath = "" to close view if still open

src/features/publications/api/useDeletePublication.ts
rubinaga: thanks, kept it simple and just made it back to "string"

src/features/publications/components/PublicationsContainer/PublicationsContainer.tsx
rubinaga: added a couple tests to cover this too, thanks for catching it

src/features/publications/components/RemovePublicationModal/RemovePublicationModal.tsx
marqode, line 36: close should set sidePath="" to close view if its open

marqode (approved):
side panel should close on remove or republish but its nonblocking if you want to merge

PR #608, opened 2026-05-07
Switch profiles side panels to new cancel design

gesquivelgaghi:
## Summary

Dani specified new standards for side panels cancel / close behavior. This PR applies these standards to the Profiles side panels, 

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/src/features/usg-profiles/components/USGProfileForm/USGProfileForm.tsx
+++ b/src/features/usg-profiles/components/USGProfileForm/USGProfileForm.tsx
@@ -1,6 +1,6 @@
 import SidePanelFormButtons from "@/components/form/SidePanelFormButtons";
 import usePageParams from "@/hooks/usePageParams";
-import type { ComponentProps, ReactNode } from "react";
+import type { ReactNode } from "react";
 import { useState, type FC } from "react";
 import { phrase } from "../../helpers";
 import type { UseUSGProfileFormProps } from "../../hooks/useUsgProfileForm";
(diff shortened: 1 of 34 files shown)


rubinaga:
@gesquivelgaghi could you please check the failing unit test? it's reproducible locally too

rubinaga:
Have you considered moving the sidepanel functions out of the `usePageParams` hook into a separate hook?

gesquivelgaghi:
> Have you considered moving the sidepanel functions out of the `usePageParams` hook into a separate hook?

@rubinaga  I thought it was fine to keep there since its fully controlled by the page params, but do you think it'd be better as like its own hook? The thing is we already have a `useSidePanel` hook, but it control side panels that are not managed by the url, so I thought that by keeping these with the pageParams hook the it'd be easier to differentiate them. If I move them out what should I name the hook?


Also the test failure is unrelated, marc already fixed it, i'll just rebase off main

src/features/usg-profiles/components/USGProfileForm/USGProfileForm.tsx
```diff
@@ -109,7 +109,7 @@ const USGProfileForm: FC<USGProfileFormProps> = ({
         submitButtonText={submitButtonText}
         hasBackButton={hasBackButton}
```
marqode: Should we remove this Back button, since it isn't a multi-step form?

marqode (changes requested):
- `USGProfileEditSidePanel` can have its back button removed
- `USGProfileDuplicateSidePanel` can have its back button removed
- `PackageProfileConstraintsAddForm` can have its back button removed
- When `PackageProfileConstraintsAddForm` succeeds it uses `popSidePath` to return to PackageProfileConstraintsEditForm, probably should go back to view or close side panel (lower priority)

src/features/usg-profiles/components/USGProfileForm/USGProfileForm.tsx
gesquivelgaghi: I addressed all the changes, the only thing is for `PackageProfileConstraintsAddForm` we do want it to pop on success so it goes back to `Edit package constraints` so I left that as is. Will push as soon as I check the tests

gesquivelgaghi:
tests passing 

image

marqode (approved):
LGTM!

PR #622, opened 2026-05-14
test: remove unnecessary mocks

rubinaga:
## Summary

Summarize the changes made in this pull request. Include any relevant context or background information that would help reviewers understand the purpose and scope of the changes.

## Release Impact

- [ ] `main` — next beta → publishes to `ppa-build`

Change type (tick one):

- [ ] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

- [ ] **Changeset added** — I have run `pnpm changeset` (or `pnpm changeset --empty` if no CHANGELOG line is warranted) and committed the resulting `.md` file. Required for PRs targeting `main` and `release/*`; enforced by the `Changeset check` workflow.

## Versioning Reminder

> [!IMPORTANT]
>


src/features/profiles/components/ProfilesList/helpers.tsx
```diff
@@ -133,11 +133,15 @@ export const getStatusColumn = (): Column<Profile> => ({
     if (isUsgProfile(profile) || isScriptProfile(profile)) {
       return getStatus(profile).icon;
     }
+
+    return undefined;
```
gesquivelgaghi: why not make this null like everything else?

src/features/profiles/index.ts
```diff
@@ -8,8 +8,10 @@ export { default as ProfilesList } from "./components/ProfilesList";
 export { default as ProfilesListActions } from "./components/ProfilesList/components/ProfilesListActions";
 export { default as RemoveProfileModal } from "./components/RemoveProfileModal";
 
+export { PROFILE_DAY_OPTIONS } from "./constants";
 export { useOpenProfileSidePanel } from "./hooks";
 
-export { ProfileTypes } from "./helpers";
+export { parseSchedule, ProfileTypes } from "./helpers";
 
+export type { ProfileDay } from "./constants";
```
gesquivelgaghi: since this is a type shouldnt it be in the `types` file?

src/features/account-creation/components/AccountCreationSelfHostedForm/AccountCreationSelfHostedForm.test.tsx
```diff
@@ -18,29 +18,18 @@ vi.mock("@/hooks/useAuth", () => ({
...
-    ...actual,
-    useLogin: () => ({
-      login: signInMock,
-      isLoggingIn: false,
-    }),
-  };
-});
-
 describe("AccountCreationSelfHostedForm", () => {
-  const user = userEvent.setup();
```
gesquivelgaghi, line 40: why is it better to have the `const user = userEvent.setup()` line in each individual test?

src/features/local-repositories/api/useGetPageLocalRepository.ts
```diff
@@ -4,7 +4,7 @@ import type { Local } from "@canonical/landscape-openapi";
 
 type GetPageLocalRepositoryReturnType =
   | {
-      repository: Local;
+      repository: Local | undefined;
```
gesquivelgaghi: what is this solving?

src/features/account-creation/components/AccountCreationSelfHostedForm/AccountCreationSelfHostedForm.test.tsx
rubinaga: I also thought declaring it once was better. But technically it transfers state from one test to the other and does not isolate, which could lead to flaky tests or annoying bugs. I don't think it necessarily affects many of our tests but probably it would be good practice to isolate it.

Can also do this:
```ts
let user;

beforeEach(() => {
    user = userEvent.setup();
})
```

What do you think?

src/features/local-repositories/api/useGetPageLocalRepository.ts
rubinaga: TICS pointed out a scenario where the returned repository's value is undefined by itself. In that case we'd be casting it as `Local`

src/features/account-creation/components/AccountCreationSelfHostedForm/AccountCreationSelfHostedForm.test.tsx
gesquivelgaghi: that makes sense. feel free to keep as is, i was just curious

src/features/local-repositories/api/useGetPageLocalRepository.ts
gesquivelgaghi: that shouldnt be the case because of the logic inside `useGetLocalRepository`, which this calls. 

But we could completely exclude undefined as an option while keeping this type-safe return type by putting the return under a condition:
```
if (repository) {
      return {
        repository: repository,
        isGettingRepository: false,
      };
    }
```

src/features/local-repositories/components/PublishLocalRepositorySidePanel/components/PublishRepositoryExistingForm/PublishRepositoryExistingForm.tsx
```diff
@@ -138,7 +138,7 @@ const PublishRepositoryExistingForm: FC<PublishRepositoryExistingFormProps> = ({
                   Automatic installation
                 </span>
                 <Tooltip
-                  message={SETTINGS_HELP_TEXT.notAutomatic}
+                  message={PUBLICATION_SETTINGS_HELP_TEXT.automaticUpgrades}
                   position="top-center"
```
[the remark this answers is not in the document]
gesquivelgaghi: these issues are actually addressed by the PR i merged today which consolidates the publication settings so we avoid inconsistent behavior like this

src/features/local-repositories/api/useGetPageLocalRepository.ts
rubinaga: I think in most places we explicitly return `data ?? null` in case something went wrong in the backend. I'd say this is more defensive programming. In this case the `if (repository)` wouldn't quite work because it would imply the `useGetLocalPageRepository` function can return `undefined`, which then is a problem for all files that consume it.

src/features/mirrors/components/AddMirrorForm/AddMirrorForm.test.tsx
```diff
@@ -7,56 +7,37 @@ import {
...
     );
 
-    await user.click(screen.getByRole("button", { name: "Add mirror" }));
-
-    expect(mockCreateMirror).toHaveBeenCalledExactlyOnceWith(
-      expect.objectContaining({
-        archiveRoot: `https://${UBUNTU_ARCHIVE_HOST}/ubuntu/`,
-      }),
-    );
+    await submitMirrorAndExpectSuccess();
```
gesquivelgaghi: These AddMirrorForm tests used to test that the fields are set correctly, but now they all just check the same thing (that it submits successfully).

rubinaga: Thanks for catching this I missed it. Changed it back to how it was

src/features/profiles/components/AddProfileButton/AddProfileButton.test.tsx
```diff
@@ -1,62 +1,68 @@
...
     ).not.toHaveClass("p-button--positive");
   });
 
-  it("opens add side panel on click", async () => {
-    renderWithProviders(<AddProfileButton />);
-
-    await userEvent.click(screen.getByRole("button", { name: /add profile/i }));
-
-    expect(openAddSidePanel).toHaveBeenCalledTimes(1);
+  it("opens add side panel for non-WSL type", async () => {
```
gesquivelgaghi: there's no difference for WSL profiles anymore since they're no longer a beta feature

src/features/publication-targets/components/EditTargetForm/EditTargetForm.test.tsx
```diff
@@ -1,25 +1,19 @@
...
-  throw new Error("Test targets must have S3 config");
-}
-if (!swiftTarget?.swift) {
-  throw new Error("Test target must have Swift config");
-}
-if (!filesystemTarget?.filesystem) {
-  throw new Error("Test target must have Filesystem config");
-}
+assert(s3TargetFull?.s3, "Test targets must have S3 config");
+assert(s3TargetMinimal?.s3, "Test targets must have S3 config");
```
gesquivelgaghi: nitpick: since these are asserted individually now, it should use singular phrasing
```suggestion
assert(s3TargetFull?.s3, "Test target must have S3 config");
assert(s3TargetMinimal?.s3, "Test target must have S3 config");
```

src/features/local-repositories/components/AddLocalRepositoryButton/AddLocalRepositoryButton.test.tsx
```diff
@@ -1,20 +1,14 @@
...
-  default: vi.fn(),
-}));
-
-const mockCreatePageParamsSetter = vi.fn();
-const mockUsePageParams = vi.mocked(usePageParams);
-
-mockUsePageParams.mockReturnValue({
-  createPageParamsSetter: mockCreatePageParamsSetter,
-} as unknown as ReturnType<typeof usePageParams>);
+const LocationDisplay = () => {
```
gesquivelgaghi, line 8: Maybe we should extract this into a reusable Test component, since we use it in so many files now. But you can just make a ticket for it if you think its out of scope for this work

src/features/profiles/components/RemoveProfileModal/RemoveProfileModal.test.tsx
```diff
@@ -3,39 +3,32 @@ import userEvent from "@testing-library/user-event";
...
-  default: vi.fn(),
-}));
-
-const mockUseProfiles = vi.mocked(useProfiles);
-
 const [baseProfile] = profiles;
 
 describe("RemoveProfileModal", () => {
-  const removeProfile = vi.fn();
-  const closeModal = vi.fn();
```
gesquivelgaghi, line 20: why are we redefining these in each test? Isnt it enough to just clear the mocks in the beforeEach, like we're doing?

src/features/local-repositories/components/AddLocalRepositoryButton/AddLocalRepositoryButton.test.tsx
rubinaga: I agree, but I think it should be a separate PR instead of in this one

src/features/profiles/components/RemoveProfileModal/RemoveProfileModal.test.tsx
rubinaga: Fair enough, either one of the other should have been used. Changed it back to how it was

src/features/profiles/components/AddProfileButton/AddProfileButton.test.tsx
```diff
@@ -1,62 +1,68 @@
...
-    mockCreatePageParamsSetter.mockReturnValue(openAddSidePanel);
-
-    mockUsePageParams.mockReturnValue({
-      createPageParamsSetter: mockCreatePageParamsSetter,
-    } as unknown as ReturnType<typeof usePageParams>);
-
-    mockUseProfiles.mockReturnValue({
-      isProfileLimitReached: false,
-    } as ReturnType<typeof useProfiles>);
+    localStorage.clear();
```
gesquivelgaghi: we dont need to clear the local storage anymore either, since its no longer used in the component

src/features/repository-profiles/components/RepositoryProfileList/RepositoryProfileList.test.tsx
```diff
@@ -64,20 +62,40 @@ describe("RepositoryProfileList", () => {
...
+
+    const profileButton = within(row).getByRole("button", {
+      name: firstProfile.title,
+    });
+    await user.click(profileButton);
 
-    expect(screen.getByTestId("location")).toHaveTextContent("sidePath=view");
+    expect(
+      within(row).getByRole("button", { name: firstProfile.title }),
+    ).toBeInTheDocument();
```
gesquivelgaghi: why are we looking for this button again when its already defined above?

src/features/repository-profiles/components/RepositoryProfileList/RepositoryProfileList.test.tsx
```diff
@@ -39,14 +34,17 @@ describe("RepositoryProfileList", () => {
       const accessGroup = accessGroups.find(
         (group) => group.name === profile.access_group,
       )?.title;
-      assert(accessGroup);
 
-      expect(await within(row).findByText(accessGroup)).toBeInTheDocument();
+      expect(accessGroup).toBeDefined();
+      expect(
+        await within(row).findByText(accessGroup ?? ""),
+      ).toBeInTheDocument();
```
gesquivelgaghi: I dont understand these changes, it feels like it was better before when we asserted accessGroup and didnt have to have the undefined check w fallback

src/tests/controllers/controller.ts
```diff
@@ -1,25 +1,78 @@
...
+  if (globalStatus) {
+    return globalStatus;
+  }
+
+  if (hasOneItem(endpointStatuses)) {
+    return endpointStatuses[0];
+  }
+
+  return DEFAULT_ENDPOINT_STATUS;
+};
```
gesquivelgaghi: seems to me like this is the exact same logic with less code being duplicated in both branches. Lmk if I'm missing something

```suggestion
  }

  const globalStatus = endpointStatuses.find(
    (status) => !normalizeEndpointPath(status.path),
  );

  if (globalStatus) {
    return globalStatus;
  }

  if (!path && hasOneItem(endpointStatuses)) {
    return endpointStatuses[0];
  }

  return DEFAULT_ENDPOINT_STATUS;
};
```

src/features/local-repositories/api/useGetPageLocalRepository.ts
gesquivelgaghi: what about doing the 

```
  if (!repository) {
    throw new Error("Local repository not found");
  }
```

here instead of in each file that uses this?

src/features/repository-profiles/components/RepositoryProfileList/RepositoryProfileList.test.tsx
rubinaga: this was done because clicking on the referenced button triggers navigation via react router, and the stale reference failed, but tbh I removed it as a line as it doesn't really add much to the test case after the button has already been clicked

src/tests/controllers/controller.test.ts
```diff
@@ -0,0 +1,49 @@
...
+  it("returns a single targeted status for legacy no-path callers", () => {
+    setEndpointStatus({ status: "error", path: "logout" });
+
+    expect(getEndpointStatus().status).toBe("error");
+    expect(shouldApplyEndpointStatus()).toBe(false);
+  });
+
+  it("normalizes endpoint paths before matching", () => {
+    expect(normalizeEndpointPath("/mirrors/")).toBe("mirrors");
+    expect(normalizeEndpointPath(undefined)).toBe("");
```
gesquivelgaghi, line 47: These tests arent being run because our vitest config excludes `"**/tests/**"`, which is why this doesnt fail even though `normalizeEndpointPath` returns `undefined` in this case, not `""`

rubinaga: removed the `**/tests/**` exclusion since it was redundant to begin with, this passes now

gesquivelgaghi:
Looks good, but you're gonna need an empty changeset before I can approve bc now any new pushes dismiss previous approvals as stale

gesquivelgaghi (approved):

src/tests/render.tsx
```diff
@@ -41,9 +43,23 @@ export const renderWithProviders = (
   options?: RenderOptions,
   routePath?: string,
   routePattern?: string,
+  additionalProviders?: AdditionalProviders,
```
ethanashaw, line 46: It would be good to combine these parameters into one object, but I think we can do that in a later change

ethanashaw (approved):

PR #629, opened 2026-05-20
Feat: Limit usage of submitButtonDisabled prop in SidePanelFormButtons

marqode:
## Summary

Resolves [WD-35398](https://warthogs.atlassian.net/browse/WD-35398)
- If a form has no validation errors but can't be submitted (eg `ImportRepositoryPackagesSidePanel.tsx`) use `notify.error` to show an informative message when they click submit, ideally without disabling the button.
- Keep disabled button for cases where validation error already appears in the UI (eg for `ImportRepositoryPackagesSidePanel.tsx`)

## Release Impact

Change type (tick one):

- [ ] Patch (fix)
- [X] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

[WD-35398]: https://warthogs.atlassian.net/browse/WD-35398?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/features/snaps/components/EditSnap/EditSnap.tsx
```diff
@@ -77,15 +78,29 @@ const EditSnap: FC<EditSnapProps> = ({ installedSnaps, type }) => {
     },
     validationSchema: Yup.object().shape(validationSchema),
     onSubmit: async (values) => {
+      let actionValue = type.toLowerCase();
+      if (type === EditSnapType.Uninstall) {
+        actionValue = "remove";
+      } else if (type === EditSnapType.Switch) {
+        actionValue = "refresh";
+      }
```
gesquivelgaghi:
```suggestion
      const getActionValue = () => {
        if (type === EditSnapType.Uninstall) {
          return "remove";
        }
        if (type === EditSnapType.Switch) {
          return "refresh";
        }
        return type.toLowerCase();
      }
```
just a suggestion that can go for all the similar logic below as well. it helps reduce the average cyclomatic complexity. could also use a `switch` instead of ifs but thats up to your preference

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
```diff
@@ -136,6 +145,10 @@ const WslInstanceInstallForm: FC = () => {
     await formik.setFieldValue("cloudInit", null);
   };
 
+  if (!hasValidInstanceId) {
+    return null;
+  }
```
gesquivelgaghi, line 154: what purpose does this serve, since we already use this flag to close the sidepanel on lines 35-38 above?

src/features/usg-profiles/components/USGProfileAddSidePanel/USGProfileAddSidePanel.tsx
gesquivelgaghi, line 1: there's a couple of validation issues in this form:

missing validation in step 2 for the unselected dropdowns

https://github.com/user-attachments/assets/bcf9021d-33b9-439a-806e-e726f2729640

missing validation for the first repeat every field for the schedule interval, and this is also an issue for the edit form, not just this add one (side note: its weird that there's 2 fields labeled "repeat every". i think we should rename the second one, since "repeat every days" doesnt even make sense)

https://github.com/user-attachments/assets/4a2f853c-c56f-4532-8a06-f63769acf4d7

src/features/usg-profiles/components/USGProfileAddSidePanel/USGProfileAddSidePanel.test.tsx
```diff
@@ -27,4 +41,26 @@ describe("USGProfileAddSidePanel", () => {
       screen.queryByRole("button", { name: "Back" }),
     ).not.toBeInTheDocument();
   });
+
+  it("should block on step 2 when benchmark fields are missing", async () => {
```
gesquivelgaghi, line 50: this test should check for formik validation as well, to avoid the current issue repeating itself. Also should include this type of test for every step of the form, since there's also similar issues with step 3

src/features/usg-profiles/components/USGProfileForm/USGProfileForm.tsx
gesquivelgaghi, line 1: here's the validation issues within the schedule block. again its the `repeat every` number missing, and also the `on` if the repeat every is set to `weeks` or `years`

https://github.com/user-attachments/assets/c55c4c64-a13f-4fcf-a8f6-ce9effe82696

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.tsx
```diff
@@ -84,7 +84,9 @@ const ImportRepositoryPackagesSidePanel: FC = () => {
     initialValues: { source: "" },
     onSubmit: handleSubmit,
     validationSchema: Yup.object().shape({
-      source: Yup.string().required("This field is required."),
+      source: Yup.string()
+        .required("This field is required.")
+        .url("Please enter a valid URL."),
```
gesquivelgaghi: the user should be able to input a local url, but with this change its not supported

Image

vs

Image

marqode: Thanks I thought that was only external for some reason. Updating in my next commit

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
marqode: good catch, removed

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.tsx
```diff
@@ -108,15 +110,40 @@ const ImportRepositoryPackagesSidePanel: FC = () => {
...
+      } else if (validationTask.count === 0) {
+        notify.error({
+          title: "No packages available",
+          message: "No packages available from the URL provided.",
+        });
+      }
+      return;
+    }
+    formik.handleSubmit();
+  };
```
gesquivelgaghi:
```suggestion
  const shouldDisableImportButton =
    !!validationTask?.error &&
    validationTask.error.code !== 4 || 
    validationTask?.done &&
    validationTask?.count === 0;

  const handleImportClick = () => {
    formik.handleSubmit();
  }; // so actually just delete this function and use formik's handleSubmit directly
```

as an explanation:
- the ```if (validationTask.error && validationTask.error.code !== 4)``` branch will never be reached since the button is disabled for that case
- i think the ```else if (validationTask.count === 0)``` should follow the same logic as the error case above (i.e, disabling the import button), since we are already giving the user the reason why its disabled (the `No packages available from the URL provided` message)
- this leaves only ```if (!canImport && validationTask?.done) return;```, but `&& validationTask?.done)` is redundant because of the second check in `canImport`. So we're left with just ```if (!canImport) return;```, but if we dig into that logic, we realize all the `!canImport` cases left are already covered by the disabled button with the suggested changes:
```
!canImport = validationTask && validationTask.done && validationTask.error?.code !== 4 && 
  (validationTask.error || validationTask.count === 0)
```
can be rewritten as 
```
!canImport = 
  validationTask?.done && // redundant here since validationTask?.error is only true if the task is done
  validationTask?.error && validationTask?.error?.code !== 4  ||
  validationTask?.error?.code !== 4 && // redundant here since its already covered by the case above
  validationTask?.done && validationTask?.count === 0
```
which is equal to
```
const shouldDisableImportButton =
  !!validationTask?.error && validationTask.error.code !== 4 || 
  validationTask?.done && validationTask?.count === 0;
```

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.tsx
gesquivelgaghi, line 143: I actually think we should also enable this button here and then add this to `handleValidate`, so instead of the button being enabled for an empty URL it'll just also trigger the formik validation
```
    if (!formik.values.source || formik.errors.source) {
      formik.setFieldTouched("source", true);
      return;
    }
```

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.tsx
```diff
@@ -133,6 +160,11 @@ const ImportRepositoryPackagesSidePanel: FC = () => {
               help={
                 "In order to upload packages, provide a URL for Landscape to fetch the packages from."
               }
+              onKeyDown={(e) => {
+                if (e.key === "Enter") {
+                  e.preventDefault();
+                }
+              }}
```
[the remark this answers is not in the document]
marqode: This is intentional- if user hits Enter in the Source URL field they may expect it to trigger Fetch packages, instead of Import (skipping fetch)

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.tsx
marqode: Since Yup doesn't have a URI validator that matches the backend's, I'll do basic validation on `{file | http}` and let the backend handle the rest

src/features/script-profiles/components/ScriptProfileForm/ScriptProfileForm.tsx
```diff
@@ -113,6 +116,14 @@ const ScriptProfileForm: FC<ScriptProfileFormProps> = ({
         return;
       }
 
+      if (associationLimitReachedRef.current) {
+        notify.error({
+          title: "Association limit reached",
+          message: `Decrease the number of associated instances below the limit of ${scriptProfileLimits?.max_num_computers} before saving.`,
+        });
+        return;
+      }
```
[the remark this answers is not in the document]
marqode: The ref is kept in sync with useLayoutEffect (not useEffect), which runs synchronously before paint. By the time the "limit reached" notification is visible to the user, the ref is already true, so there's no window in which a submit could proceed incorrectly. The pattern is intentional — the ref is used to avoid adding isAssociationLimitReached to Formik's closure, since that would cause the [onSubmit](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-browser/workbench/workbench.html) config object to change on every render.

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
```diff
@@ -28,6 +28,15 @@ const WslInstanceInstallForm: FC = () => {
...
   const openActivityDetails = useOpenActivityDetailsPanel();
+  const parsedInstanceId = Number(instanceId);
+  const hasValidInstanceId =
+    Number.isInteger(parsedInstanceId) && parsedInstanceId > 0;
+
+  useEffect(() => {
+    if (!hasValidInstanceId) {
+      closeSidePanel();
+    }
+  }, [closeSidePanel, hasValidInstanceId]);
```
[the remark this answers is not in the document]
marqode: The form is gated by `if (!hasValidInstanceId) { return null; }` before the `return (<Form ...>)`, so it doesn't render (and the submit button is unreachable) when the ID is missing or invalid. The `onSubmit` guard is a second line of defense. No change needed here.

.changeset/short-webs-sneeze.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": minor
+---
+
+Limit usage of submitButtonDisabled prop from SidePanelFormButtons. Most forms no longer disable submit based on validity and keep the button enabled except while loading, though some flows may still disable it after validation failures (for example, Import Repository Packages).
```
gesquivelgaghi:
```suggestion
Limit usage of disabled submit buttons in side panel forms. Most forms now only disable the submit button while loading, though some flows may still disable it after validation failures (for example, importing repository packages).
```
since changesets are meant to be user facing i think we should avoid using codebase terminology and instead describe the UI in simple terms

src/features/packages/components/PackagesInstallForm/PackagesInstallForm.tsx
```diff
@@ -30,6 +30,14 @@ const PackagesInstallForm: FC = () => {
   } = packagesActionQuery;
 
   const handleSubmit = async () => {
+    if (selected.length === 0) {
```
gesquivelgaghi, line 35: should this be that caution/warning on the field itself instead, like discussed with Dani for those other fields? seems like its a similar usage of having no changes/input to a form

src/features/snaps/components/InstallSnaps/InstallSnaps.tsx
```diff
@@ -25,6 +25,14 @@ const InstallSnaps: FC = () => {
   const instanceId = Number(urlInstanceId);
 
   const handleSubmit = async () => {
+    if (selectedSnaps.length === 0) {
```
gesquivelgaghi, line 34: same question here, should this be an input caution warning instead?

src/features/ubuntupro/components/TokenFormBase/TokenFormBase.tsx
```diff
@@ -150,7 +150,6 @@ const TokenFormBase: FC<TokenFormBaseProps> = ({
 
         <SidePanelFormButtons
           submitButtonText={submitButtonText}
-          submitButtonDisabled={!formik.isValid || !formik.dirty}
```
gesquivelgaghi, line 153: this is definitely outside the scope of this PR but I think we should figure out a consistent way to deal with submission when forms aren't dirty. I'll make a ticket for it 

update: here https://warthogs.atlassian.net/browse/LNDENG-4783

src/pages/dashboard/settings/administrators/tabs/administrators/AdministratorList/AdministratorList.tsx
```diff
@@ -44,7 +44,10 @@ const AdministratorList: FC<AdministratorListProps> = ({
   const handleAdministratorClick = (administrator: Administrator) => {
     setSidePanelContent(
       administrator.name,
-      <EditAdministratorForm administrator={administrator} />,
+      <EditAdministratorForm
+        key={administrator.id}
```
gesquivelgaghi, line 48: what is this for? i thought the `key` attribute was only for elements in lists

src/features/autoinstall-files/components/AutoinstallFileForm/AutoinstallFileForm.test.tsx
```diff
@@ -125,17 +132,11 @@ describe("AutoinstallFileForm", () => {
       <AutoinstallFileForm {...createAutoinstallFileWithContentsProps()} />,
     );
 
-    expect(screen.getByRole("button", { name: "Add" })).toHaveAttribute(
-      "aria-disabled",
-      "true",
-    );
+    expect(screen.getByRole("button", { name: "Add" })).toBeEnabled();
```
gesquivelgaghi:
```suggestion
    expect(screen.getByRole("button", { name: "Add" })).not.toHaveAttribute(
      "aria-disabled",
      "true",
    );
```
the issue is `expect(screen.getByRole("button", { name: "Add" })).toBeEnabled();` passes even if `aria-disabled="true"`, so its not actually testing anything, since we never actually disable our buttons, only aria-disable them.

There's a lot more usages of this across the PR, I think if you just tell copilot to change them it'll be easier than either of us trying to not miss any

rubinaga (dismissed):
LGTM

gesquivelgaghi (changes requested):
not done but here are some comments I have so far. most are nitpicky or possibly out of scope, but I think the `toBeEnabled` checks need to be fixed

src/features/instances/components/DistributionUpgrades/DistributionUpgrades.tsx
```diff
@@ -177,7 +189,13 @@ const DistributionUpgrades: FC<DistributionUpgradesProps> = ({
   }
 
   return (
-    <Form noValidate onSubmit={openModal}>
+    <Form
+      noValidate
+      onSubmit={(e) => {
+        e.preventDefault();
+        handleOpenModal();
+      }}
```
gesquivelgaghi: why do we need both `onSubmit`s?

src/pages/dashboard/settings/administrators/tabs/administrators/EditAdministratorForm/EditAdministratorForm.tsx
```diff
@@ -87,9 +91,7 @@ const EditAdministratorForm: FC<EditAdministratorFormProps> = ({
   const formik = useFormik({
     initialValues: { roles: [] as string[] },
     validationSchema: Yup.object().shape({
-      roles: Yup.array()
-        .of(Yup.string())
-        .min(1, "At least one role is required"),
```
gesquivelgaghi, line 92: why dont we want at least 1 role anymore?

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
gesquivelgaghi: did this get re-added for a reason?

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
```diff
@@ -78,6 +87,15 @@ const WslInstanceInstallForm: FC = () => {
...
     }),
     onSubmit: async (values) => {
+      if (!hasValidInstanceId) {
+        notify.error({
+          title: "Invalid parent instance",
+          message: "Unable to determine the parent instance. Please try again.",
+        });
+        closeSidePanel();
+        return;
+      }
```
gesquivelgaghi, line 87: does this even show before lines 35-38? seems like these lines conflict with each other

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.tsx
```diff
@@ -105,15 +112,15 @@ const ImportRepositoryPackagesSidePanel: FC = () => {
...
-    (validationTask?.status === "succeeded" && !!validationTask.count);
-
   const packagesCount =
     validationTask && validationTask.count > 0
       ? pluralize(validationTask.count, ["package"], "exact")
       : "packages";
 
+  const shouldDisableImportButton =
+    (!!validationTask?.error && validationTask.error.code !== 4) ||
+    (validationTask?.done && validationTask?.count === 0);
```
gesquivelgaghi:
```suggestion
    (validationTask?.status === "succeeded" && validationTask?.count === 0);
```

otherwise it will also disable the button for the timeout error

src/features/usg-profiles/components/USGProfileForm/USGProfileForm.tsx
```diff
@@ -17,6 +18,18 @@ interface USGProfileFormProps extends UseUSGProfileFormProps {
...
+  "title",
+  "benchmark",
+  "mode",
+  "start_type",
+  "start_date",
+  "every",
+  "end_date",
+  "deliver_delay_window",
+  "restart_deliver_delay",
+];
```
gesquivelgaghi, line 33: we're missing validation in the required field `On` that only shows conditionally if `week(s)` or `year(s)` is selected for `Repeat every`
Image

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.test.tsx
```diff
@@ -176,4 +261,243 @@ describe("ImportRepositoryPackagesSidePanel", () => {
...
+    renderComponent();
+
+    const input = await screen.findByLabelText(/source url/i);
+    await user.type(input, "not a url");
+    await user.tab();
+
+    await waitFor(() => {
+      expect(screen.getByText(/please enter a valid url/i)).toBeInTheDocument();
+    });
+  });
```
gesquivelgaghi: this is a duplicate of the test in lines 211-230

src/features/local-repositories/components/ImportRepositoryPackagesSidePanel/ImportRepositoryPackagesSidePanel.test.tsx
```diff
@@ -165,6 +196,60 @@ describe("ImportRepositoryPackagesSidePanel", () => {
...
+    renderComponent();
+
+    const input = await screen.findByLabelText(/source url/i);
+    await user.type(input, "not a url");
+
+    const importButton = screen.getByRole("button", {
+      name: /import packages/i,
+    });
+    expect(importButton).toBeEnabled();
+  });
```
gesquivelgaghi: this test is already covered by the test in lines 232-251

src/pages/dashboard/settings/administrators/tabs/administrators/AdministratorList/AdministratorList.tsx
marqode: Without it the form's contents can become stale when you click a new admin in the list


https://github.com/user-attachments/assets/39fa7723-8c85-42ba-bafd-b609d235aff8

src/features/instances/components/DistributionUpgrades/DistributionUpgrades.tsx
marqode: see [this copilot review comment](https://github.com/canonical/landscape-ui/pull/629#discussion_r3276915375)- it's a little outside the scope of this PR but seemed like a minor enough change to get the form behavior we want, ie supporting Enter-to-submit

src/pages/dashboard/settings/administrators/tabs/administrators/EditAdministratorForm/EditAdministratorForm.tsx
marqode: You're right we should keep that check- added it back.

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
marqode: This was just an extra fallback in case the instance id was somehow undefined when they go to submit. In practice that would always trigger closing the form from the `useEffect` so this should be safe to remove.

src/features/autoinstall-files/components/AutoinstallFileForm/AutoinstallFileForm.test.tsx
marqode: Thanks for pointing this out. I see a lot of uses of `.toBeEnabled()` throughout the codebase, do you think we need a follow-up ticket to review / replace those?

src/features/packages/components/PackagesInstallForm/PackagesInstallForm.tsx
marqode: This seems different to me than the "no changes" case- admittedly it's pretty subjective, but failing to install packages feels more like an error where failing to save with no changes feels like it warrants a warning. I asked Dani in [that thread](https://chat.canonical.com/canonical/pl/9mybj4szrpy6xqzhc44x8yzjre) for his opinion.

marqode:
@gesquivelgaghi thanks for the thorough review! I think I got everything in my most recent commit but let me know if I missed anything or you have any questions.

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
marqode: nope, I'm not sure why the removal commit didn't stick. It's gone now

src/features/autoinstall-files/components/AutoinstallFileForm/AutoinstallFileForm.test.tsx
gesquivelgaghi: Yeah, that seems worth doing

src/pages/dashboard/settings/administrators/tabs/administrators/AdministratorList/AdministratorList.tsx
gesquivelgaghi: oh interesting

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
```diff
@@ -28,6 +28,15 @@ const WslInstanceInstallForm: FC = () => {
   const { closeSidePanel } = useSidePanel();
   const { notify } = useNotify();
   const openActivityDetails = useOpenActivityDetailsPanel();
+  const parsedInstanceId = Number(instanceId);
+  const hasValidInstanceId =
+    Number.isInteger(parsedInstanceId) && parsedInstanceId > 0;
+
+  useEffect(() => {
+    if (!hasValidInstanceId) {
+      closeSidePanel();
```
gesquivelgaghi: to be honest maybe when this is the case instead of closing the side panel we should show an error/empty state. I recently had a discussion with Dani about something kind of similar to this: https://chat.canonical.com/canonical/pl/q5uosjgkip8jzqjfr1b4uxcwqo

src/features/instances/components/DistributionUpgrades/DistributionUpgrades.tsx
gesquivelgaghi: i tried testing this but its not working for me. Were you able to test it successfully?

src/features/autoinstall-files/components/AutoinstallFileForm/AutoinstallFileForm.test.tsx
marqode: https://warthogs.atlassian.net/browse/LNDENG-4790

src/features/wsl/components/WslInstanceInstallForm/WslInstanceInstallForm.tsx
marqode: I added an empty state for this and corresponding test, but I'm not sure it'll ever appear in the UI- this form doesn't use the side panel `pageParams` so there isn't a URL to open the side panel for an invalid instance ID (user sees empty state with no side panel instead)

Still, probably better to have the empty state fallback in the side panel in case we add sidePath routing at some point.

image

gesquivelgaghi: oh i see. Yeah, I do think the goal is to eventually have all side panels use the URL routing, so thats good

src/features/instances/components/DistributionUpgrades/DistributionUpgrades.tsx
marqode: Turns out I was too optimistic here.. our side panels don't trap focus so Enter actually triggers the last UI element selected before the panel was open. Since none of our other side panel forms support Enter-to-submit (except for text inputs) I removed this to align with that behavior.

gesquivelgaghi: yeah, we really need to fix the focus issue tbh

marqode: looks like there's already a ticket: LNDENG-4564

gesquivelgaghi (dismissed):

gesquivelgaghi (dismissed):

gesquivelgaghi (dismissed):

marqode:
@rubinaga @gesquivelgaghi thanks for the multiple rounds of reviews. This should be (actually) good to go at this point, I just made the change Dani requested to show "No changes" notifications at the top of forms instead of on the fields, to be consistent with the no-packages and no-snaps form behavior
image
image

gesquivelgaghi (dismissed):

src/features/autoinstall-files/components/AutoinstallFileForm/AutoinstallFileForm.test.tsx
gesquivelgaghi, line 1: actually do we want tests to cover the no changes notifications?

marqode: Good point, I'll add a test case

gesquivelgaghi (approved):

rubinaga (approved):
LGTM

PR #631, opened 2026-05-21
Feat: Containerized integration testing with landscape-packaging

marqode:
## Summary

This PR adds a github action to test landscape-ui against the backend services in landscape-server and landscape-go. The integration-test CI action includes two test suites:
- `API Contract` tests run first, to check for API drift before rendering any UI (fast)
- `e2e selfHosted / SaaS` tests run next in a playwright browser session (slower)

This action currently takes 10-14m to run 15-20 tests, much of which is spent on the docker build and data seeding steps. Build time could be reduced by using a self-hosted runner with caching.
- `API contract tests` (no browser session) take 1 min 45s
- `Run e2e` selfHosted and Saas steps combined take about 4min 30s

Note: this was originally a PR from [my repo fork](https://github.com/marqode/landscape-ui) but I merged the fork into this branch to see if that helps checks to pass.

Resolves [LNDENG-4695](https://warthogs.atlassian.net/browse/LNDENG-4695)

## Release Impact

Change type (tick one):

- [ ] Patch (fix)
- [X] Minor (feature)
- [ ] Major (breaking)

> The change-type label is informational and only affects how the entry
is rendered in the CHANGELOG. The actual version is computed from the branch and CalVer cycle — `pnpm changeset`'s `patch`/`minor` choice does not influence it.

## Checklist

## Versioning Reminder

> [!IMPORTANT]
> Landscape UI uses **CalVer** (`YY.MM.Point.Build`). Version derivation
by branch:
>

[LNDENG-4695]: https://warthogs.atlassian.net/browse/LNDENG-4695?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,372 @@
...
+          echo "landscape-debarchive database ready"
+
+      - name: Start backend stack
+        working-directory: .landscape-packaging/docker/ui-dev
+        run: |
+          _uid=$(id -u)
+          _gid=$(id -g)
+          _uname=$(whoami)
+          env UID="$_uid" GID="$_gid" UNAME="$_uname" \
+            LANDSCAPE_BOOTSTRAP_SCHEMA_ARGS="--with-computers --with-free-disk-space --with-free-memory-and-swap --with-load-averages --with-temperatures --with-network-traffic --with-active-processes --with-packages --with-package-activities --with-script-activities --with-users-and-groups --with-cpu-usage --with-ceph-usage --with-compute-usage --with-swift-usage --with-user-and-group-activities --with-custom-graph --with-scripts --with-account-password foo" \
```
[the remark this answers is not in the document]
marqode: this is intentional- account-password flag sets the password for instance updates, not login

e2e/docker-stack/api/local-repositories.spec.ts
```diff
@@ -0,0 +1,57 @@
...
+}
+
+test.describe("Local Repositories API Contract", () => {
+  const testRepoName = `test-local-repo-${Date.now()}`;
+  let createdRepoName: string | undefined;
+
+  test.afterAll(async ({ request }) => {
+    if (createdRepoName) {
+      await request.delete(`/v1beta1/${createdRepoName}`);
+    }
```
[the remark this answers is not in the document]
marqode: JWT included manually

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,385 @@
...
+      # actions/checkout only injects HTTPS credentials; landscape-packaging submodules
+      # use SSH URLs so the credential would be silently ignored. persist-credentials:
+      # false avoids a stale credential interfering with the rewrite step.
+      - name: Checkout landscape-packaging (top-level only)
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          repository: canonical/landscape-packaging
+          token: ${{ steps.get_token.outputs.token }}
+          ref: ${{ inputs.packaging_ref || 'main' }}
+          submodules: false
```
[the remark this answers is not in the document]
marqode: Risk is minimal: on every automatic trigger it falls back to 'main', so it cannot affect PR/push/nightly behavior. Leave input for custom branch specification

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,384 @@
...
+            docker compose \
+              -p "ls-integration-${{ github.run_id }}" \
+              -f compose.yaml -f compose.ci-override.yaml \
+              up -d debarchive-seeder
+
+          echo "Waiting for debarchive-seeder to finish..."
+          until docker inspect landscape-debarchive-seeder &>/dev/null; do sleep 2; done
+          EXIT=$(docker wait landscape-debarchive-seeder)
+          if [ "$EXIT" != "0" ]; then
+            echo "::warning::Debarchive seeder exited with code $EXIT (non-blocking — seeder may have partial data or use an older API version)"
```
rubinaga: Could we make this a hard failure instead of a warning?

The downstream UI tests assert that the seeded mirrors/targets exist, so if `debarchive-seeder` exits non-zero, continuing here will likely produce later Playwright timeouts like "row not visible. That makes the failure look like a UI/test issue instead of a setup/data issue.

Failing loudly here, or adding a direct seeded-data precondition with a clear error message, would make this much easier to debug and more reliable as a PR gate

e2e/docker-stack/api/local-repositories.spec.ts
```diff
@@ -0,0 +1,72 @@
...
+import { getAuthToken } from "../helpers/auth";
+
+function validateLocalShape(repo: unknown): asserts repo is Local {
+  expect(repo).not.toBeNull();
+  expect(typeof repo).toBe("object");
+  expect(repo).toHaveProperty("name");
+  expect((repo as { name?: unknown }).name).toEqual(expect.any(String));
+}
+
+test.describe("Local Repositories API Contract", () => {
```
rubinaga, line 12: This test flow has retry-sensitive shared state: `createdRepoName` is set by the POST test and then required by the GET/list test.

Since the config enables `retries: 1`, if the GET/list test flakes, Playwright may retry that test in a fresh worker where the POST test has not run, leaving `createdRepoName` unset. The retry then cant actually validate the intended behavior.

Could we combine the create/list/delete flow into one test using `test.step` + `try/finally`, or mark this describe block serial so the setup step reruns together with the dependent assertion?

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,384 @@
...
+          node-version: 24
+          cache: pnpm
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Install Playwright browsers
+        run: pnpm exec playwright install --with-deps chromium
+
+      - name: Run API Contract tests
```
rubinaga, line 336: This currently runs three separate Playwright commands/configs, and each config starts its own Vite `webServer` and runs the same `globalSetup`.

That means the backend check, browser login, storage-state generation, and archive warm-up are repeated for API-contract, self-hosted UI, and SaaS UI runs.

Could we consider consolidating this into one Playwright config with projects, using a setup project/dependency for the shared login + warm-up? Not necessarily a blocker for this PR but something to consider

e2e/docker-stack/playwright.integration.config.ts
```diff
@@ -0,0 +1,49 @@
...
+    cwd: "../../",
+    command: "vite --mode e2e.selfHosted",
+    url: BASE_URL,
+    reuseExistingServer: false,
+    timeout: 120_000,
+    stdout: "pipe",
+    stderr: "pipe",
+  },
+
+  use: {
```
rubinaga, line 44: Most self-hosted UI specs need the authenticated storage state, but this config does not set it as a default. Could we either set authenticated `storageState` as the default and explicitly opt the login spec out, or split authenticated/unauthenticated tests into separate projects?

package.json
```diff
@@ -10,6 +10,8 @@
     "dev": "vite",
     "build": "pnpm run lint && pnpm run tcm:run && tsc && vite build",
     "build:e2e": "pnpm run build --mode e2e",
+    "build:e2e.selfHosted": "pnpm run build --mode e2e.selfHosted",
```
rubinaga: Are `build:e2e.selfHosted` and `build:e2e.saas` intended to be used somewhere?

.gitignore
```diff
@@ -33,5 +33,13 @@ src/**/*.module.scss.d.ts
 *.sw?
 /test-results/
 /playwright-report/
+/playwright-integration-report/
+/playwright-integration-saas-report/
+/playwright-api-contract-report/
 /playwright/
+/docs/superpowers
```
rubinaga: could we remove this or do you think it's good to keep in there? seems more like a local artifact

rubinaga (changes requested):
Thanks for this Marc, I added some comments let me know what you think

.github/workflows/integration-tests.yml
marqode: My main goal when the seeder was still flaky was not to block the rest of the test runs, but you're right that the silent failure would be hard to debug when debarchive tests fail later on. I think either failing loudly here (simplest) or setting a flag not to run debarchive tests would be a good fix.

e2e/docker-stack/api/local-repositories.spec.ts
marqode: Good catch, refactored into one lifecycle test

.github/workflows/integration-tests.yml
marqode: At least two are necessary for saas vs. self-hosted. Combining API contract + self-hosted UI into the same config would help optimize a bit, but I'd rather wait til we decide on the testing structure / use cases before making this change.

e2e/docker-stack/playwright.integration.config.ts
marqode: Good call, added default auth state + login opt-out.

package.json
marqode: I had added them for local testing convenience but I don't think it's worth bloating `package.json` and the README doesn't reference them. Removed.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,385 @@
...
+              up -d \
+              postgresql rabbitmq rsyslog
+
+      - name: Create debarchive database
+        run: |
+          echo "Waiting for PostgreSQL to be ready..."
+          timeout 60 bash -c \
+            'until docker exec landscape-postgres pg_isready -U postgres -q; do sleep 2; done'
+          docker exec landscape-postgres createdb -U postgres landscape-debarchive || true
+          echo "landscape-debarchive database ready"
```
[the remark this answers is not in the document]
marqode: container_name: landscape-postgres hardcoded on line 60

e2e/docker-stack/README.md
```diff
@@ -0,0 +1,143 @@
...
+- `canonical/landscape-packaging` cloned with submodules (e.g., `~/landscape-packaging`)
+- `pnpm` + Node 24
+
+### 1. Start the backend with sample data
+
+Start the stack with the schema bootstrapping arguments matching CI:
+
+```bash
+# In your landscape-packaging/docker/ui-dev/ directory
+LANDSCAPE_BOOTSTRAP_SCHEMA_ARGS="--with-computers --with-free-disk-space --with-free-memory-and-swap --with-load-averages --with-temperatures --with-network-traffic --with-active-processes --with-packages --with-package-activities --with-script-activities --with-users-and-groups --with-cpu-usage --with-ceph-usage --with-compute-usage --with-swift-usage --with-user-and-group-activities --with-custom-graph --with-scripts --with-account-password foo" make up
```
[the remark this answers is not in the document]
marqode: This is not the user password referenced by CI docs

.github/workflows/integration-tests.yml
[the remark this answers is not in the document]
marqode: proposed fix introduces fragility, real risk here is low:
- The step is immediately preceded by a pg_isready poll, so connectivity and service availability are already verified.
- The container runs a fresh PostgreSQL with -U postgres (superuser), so permission errors are unlikely.
- In CI, the landscape-debarchive database will never pre-exist (fresh container each run), so the || true isn't even needed — it was added for idempotency/safety but doesn't fire in practice.

e2e/docker-stack/playwright.api-contract.config.ts
```diff
@@ -0,0 +1,43 @@
...
+
+const BASE_URL = "http://localhost:5173";
+
+export default defineConfig({
+  testDir: "./api",
+  testMatch: "**/*.spec.ts",
+
+  workers: 1,
+  fullyParallel: false,
+  retries: 1,
```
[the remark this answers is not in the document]
marqode: handled in playwright afterEach / afterAll. Wait and fix if it becomes an issue, but seems unlikely.

e2e/docker-stack/playwright.integration.config.ts
```diff
@@ -0,0 +1,52 @@
...
+  testMatch: "**/*.integration.spec.ts",
+  // SaaS-mode tests use a different Vite server mode and must run separately.
+  // Exclude them here so self-hosted mode doesn't try to run them (they would
+  // fail because VITE_SELF_HOSTED_ENV=true and routes don't redirect).
+  testIgnore: "**/*.saas.integration.spec.ts",
+
+  // Shared backend is mutable; keep workers=1 to avoid backend races.
+  workers: 1,
+  fullyParallel: false,
+  retries: 1,
```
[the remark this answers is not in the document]
marqode: handled in playwright afterEach / afterAll. Wait and fix if it becomes an issue, but seems unlikely.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,385 @@
...
+  workflow_dispatch:
+    inputs:
+      packaging_ref:
+        description: "landscape-packaging branch/tag/SHA to test against. Defaults to main. Controls which pinned versions of landscape-go and landscape-server are used."
+        default: "main"
+        required: false
+
+permissions:
+  contents: read
+
```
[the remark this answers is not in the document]
marqode: this permission isn't actually required for playwright artifacts

yurii-vasyliev (dismissed):
Really solid work — the workflow is unusually well-documented, and the credential handling (short-lived App tokens, trap-unset of the insteadOf rewrites, fork-PR gate) is careful. Verified the current head against the open review threads:

- The retry-sensitive shared state in `local-repositories.spec.ts` is resolved — the lifecycle is now a single test with `test.step`s and try/finally cleanup, so a retry re-runs the whole flow.
- Authenticated `storageState` is now the config default, with the login spec explicitly opting out.
- @rubinaga's point about the three configs each booting their own Vite server and re-running `globalSetup` still stands, but it's non-blocking for me: the SaaS run genuinely needs its own Vite mode, and the warm-up is a no-op once warm. That said, the API-contract and self-hosted configs use the same `e2e.selfHosted` mode — merging them into one config with two projects would save a webServer boot + login per run. Fine as a follow-up.

Two nits, also non-blocking:

1. `global-setup.ts` soft-fails the archive warm-up with a `console.warn` — if the 90s warm-up times out, the mirror CRUD tests will flake on a disabled Distribution select far from the root cause. Same reasoning that made the seeder a hard failure applies here.
2. The three test steps re-pass `CI_ADMIN_EMAIL`/`CI_ADMIN_PASSWORD` via step-level `env:` that just reference the job-level env, which already propagates — can be dropped.

e2e/docker-stack/api/local-repositories.spec.ts
```diff
@@ -0,0 +1,76 @@
...
+
+        const found = body.locals.find(
+          (repo: Local) => repo.name === createdRepoName,
+        );
+        expect(found).toBeDefined();
+      });
+    } finally {
+      if (createdRepoName) {
+        await test.step("delete created local repository", async () => {
+          await request.delete(`/v1beta1/${createdRepoName}`, {
```
rubinaga: Could you assert the `DELETE` response here? A `4xx/5xx` response would not fail this POST/LIST/DELETE lifecycle test, so the test could pass while deletion is broken.

rubinaga (approved):
LGTM, I just had 1 minor thing I pointed out

e2e/docker-stack/api/local-repositories.spec.ts
marqode: good catch, done

rubinaga (dismissed):
LGTM

e2e/docker-stack/helpers/ui.ts
```diff
@@ -0,0 +1,8 @@
+import type { Page } from "@playwright/test";
+
+/** Suppress the first-run welcome modal that otherwise intercepts page clicks. */
+export async function dismissWelcomePopup(page: Page): Promise<void> {
```
yurii-vasyliev: Dead code now — nothing imports `dismissWelcomePopup` since the dismissal moved into global-setup. Can be deleted.

e2e/docker-stack/global-setup.ts
```diff
@@ -0,0 +1,169 @@
...
+              lastBodySummary = String(error);
+              await new Promise((resolve) =>
+                setTimeout(resolve, ARCHIVE_WARM_POLL_MS),
+              );
+              continue;
+            }
+
+            lastStatus = res.status();
+            const rawBody = await res.text();
+            lastBodySummary = rawBody.slice(0, SERVER_ERROR);
```
yurii-vasyliev: `SERVER_ERROR` (HTTP 500) is being used as a character count here. Rename to e.g. `BODY_PREVIEW_CHARS`.

e2e/docker-stack/ui/mirrors/mirrors-crud.integration.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+    await page.waitForLoadState("networkidle");
+
+    // Find the row for our (now-renamed) mirror.
+    const mirrorRow = page
+      .getByRole("row")
+      .filter({ hasText: mirrorDisplayName });
+    await expect(mirrorRow).toBeVisible({ timeout: 15_000 });
+
+    // Open the actions menu.
+    await mirrorRow.getByRole("button").last().click();
```
yurii-vasyliev: The edit test above uses the accessible name (`"${mirrorDisplayName} mirror actions"`); `.last()` is brittle — use the same selector here.

e2e/docker-stack/ui/mirrors/mirrors-crud.integration.spec.ts
```diff
@@ -0,0 +1,277 @@
...
+ * ───────────
+ * The debarchive-seeder creates "Ubuntu Noble Main" and "Ubuntu Jammy Main"
+ * before tests run. This test creates its own uniquely-named mirror to avoid
+ * collisions with seeded data.
+ *
+ * AUTH
+ * ────
+ * The debarchive service uses JWT auth managed transparently by
+ * useFetchDebArchive. The storageState from global-setup is sufficient for
+ * all UI interactions. Direct API calls in afterAll use the v2 /me endpoint
```
yurii-vasyliev: Stale after the auth refactor — tokens now come from `POST /api/v2/login` via `getAuthToken`, not `/me`. Same note in the publication-targets-crud and preferences spec headers.

.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,430 @@
...
+                retries: 12
+                start_period: 30s
+          EOF
+          echo "=== compose.ci-override.yaml ==="
+          cat .landscape-packaging/docker/ui-dev/compose.ci-override.yaml
+
+      # Builds a production-like debarchive image from debarchive/docker/Dockerfile.e2e
+      # (multi-stage: golang:1.26 builder → debian:bookworm-slim runtime).
+      # No air, no dlv, no ptrace. The binary is at /root/server; CMD ["./server"].
+      # vendor/modules.txt was patched above so the COPY . . layer is already correct.
```
yurii-vasyliev: Nothing patches `vendor/modules.txt` anymore (the vendor step just runs `go mod vendor`) — drop this line.

e2e/docker-stack/ui/instances/instances.integration.spec.ts
```diff
@@ -0,0 +1,22 @@
+import { test, expect } from "@playwright/test";
+
+// Re-use the session authenticated in global-setup — no login step needed.
+test.use({ storageState: "e2e/docker-stack/.auth/state.json" });
```
yurii-vasyliev: Optional: both UI configs now set `storageState` as the default, so the per-spec `test.use({ storageState })` is redundant everywhere except the login spec's opt-out. Same for the README templates.

docs/integration-testing.md
```diff
@@ -0,0 +1,253 @@
...
+| `secrets.LANDSCAPE_PACKAGER_PRIVATE_KEY` | Secret              | GitHub App private key (PEM)                                            |
+| `secrets.LANDSCAPE_PROTO_TOKEN`          | Secret              | Fine-grained PAT — `Contents: Read` on `canonical/landscape-proto` only |
+
+The GitHub App must be installed on `canonical/landscape-packaging`, `canonical/landscape-go`, and `canonical/landscape-server`.
+
+> **`LANDSCAPE_PROTO_TOKEN` migration path:** `landscape-proto` is not yet covered by the App
+> installation. A fine-grained PAT is used as a temporary workaround. Once the App is
+> installed on `landscape-proto`, add it to the `repositories:` list in the
+> `create-github-app-token` step and remove the `git config url.insteadOf` line (~4 lines).
+> See [debarchive-feature-context.md](debarchive-feature-context.md) for the full credential chain.
```
yurii-vasyliev: `debarchive-feature-context.md` doesn't exist on this branch (linked 5× in this doc). Either add it or drop the references.

docs/integration-testing.md
```diff
@@ -0,0 +1,253 @@
...
+| Step                                              | Duration       |
+| ------------------------------------------------- | -------------- |
+| Initialise submodules                             | 10 s           |
+| **Start backend stack** (cold Docker image build) | **4 min 32 s** |
+| Wait for API ready (after builder exits)          | 10 s           |
+| Seed admin account                                | 4 s            |
+| Install Playwright browsers                       | 2 min 49 s     |
+| Run integration tests                             | ~30 s          |
+| **Total**                                         | **~8 min**     |
+
```
yurii-vasyliev, line 79: Table is stale vs the current run: no "Seed admin account" step anymore, Playwright install is ~25 s, three test suites take ~2.5 min, total ~10 min.

docs/integration-testing.md
```diff
@@ -0,0 +1,253 @@
...
+```bash
+curl -sf http://localhost:9091/api/v2/login/methods
+curl -sf http://localhost:8080/
+```
+
+### 2. Seed the admin account (one-time per fresh stack)
+
+```bash
+LANDSCAPE_BOOTSTRAP_SCHEMA_ARGS="--with-computers --with-free-disk-space --with-free-memory-and-swap --with-load-averages --with-temperatures --with-network-traffic --with-active-processes --with-packages --with-package-activities --with-script-activities --with-users-and-groups --with-cpu-usage --with-ceph-usage --with-compute-usage --with-swift-usage --with-user-and-group-activities --with-custom-graph --with-scripts --with-account-password foo" make up
```
yurii-vasyliev: This runs `make up` twice (plain first, then with `LANDSCAPE_BOOTSTRAP_SCHEMA_ARGS`). The builder bootstraps the schema on the first `make up`, so the sample data would be skipped. The e2e README shows the correct single command — make the two agree.

docs/integration-testing.md
```diff
@@ -0,0 +1,253 @@
...
+| `*.saas.integration.spec.ts` naming                                                        | Excluded from self-hosted config via `testIgnore`; picked up only by `playwright.integration.saas.config.ts`. The naming convention is self-documenting and requires no per-test config.                                                                                                                                           |
+| Relative API URLs in `.env.e2e`                                                            | `/api/v2/`, `/api/`, `/v1beta1/` route through the Vite proxy. `VITE_API_PROXY_TARGET` and `VITE_API_DEBARCHIVE_PROXY_TARGET` configure the targets.                                                                                                                                                                               |
```
yurii-vasyliev: `/v1beta1/` → the env files use `/v1/`.

docs/integration-testing.md
```diff
@@ -0,0 +1,253 @@
...
+  approach requires explicit workarounds for every dev-specific field that bleeds through from
+  upstream (`command`, `working_dir`, `healthcheck`, `pull_policy`, `GO_DOTENV`, `GOFLAGS`).
+  A standalone file eliminates this class of issue entirely.
+  Alternative: upstream PR to `landscape-packaging` adding a `--profile ci` compose variant.
+  See [debarchive-feature-context.md](debarchive-feature-context.md) for the full option
+  analysis (Option A / B / C).
+
+  **Gate: seeder stability.** Do not migrate to a standalone file until all of the following
+  are true:
+  1. The debarchive seeder exits 0 in CI without the non-blocking workaround — meaning the
```
yurii-vasyliev: The `::warning::` seeder bypass described in gates 1–2 no longer exists — the seeder is a hard `exit 1` now. Trim this section.

yurii-vasyliev (commented):
Re-reviewed the current head — everything from the earlier rounds is addressed and CI is green. A few small cleanups inline (mostly stale comments/docs after the auth + seeder changes); happy to approve once they're in.

FWIW the open Copilot findings on `**.md`, `inputs.packaging_ref`, and the contract suite needing global-setup are false positives — no action needed there.

yurii-vasyliev (approved):
LGTM, thanks!

rubinaga (approved):
LGTM

PR #639, opened 2026-05-25
Update dependencies

ethanashaw:


RyanLoi98 (approved):
These just look like the versioning updates as you mentioned. So, as long as these don't have any issues with backwards compatibility (which I don't expect there to be), I don't foresee any issues. LGTM!

Thanks for taking the time to fix the issues!

gesquivelgaghi (approved):

PR #647, opened 2026-06-01
Update success notifications for debarchive

gesquivelgaghi:
## Summary

- Make the phrasing of the debarchive success notifications consistent
- Add `Update mirror` action to the notification after creating a mirror

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/src/features/mirrors/components/MirrorDetails/MirrorDetails.tsx
+++ b/src/features/mirrors/components/MirrorDetails/MirrorDetails.tsx
@@ -1,4 +1,4 @@
-import { useState, type FC } from "react";
+import { useEffect, useState, type FC } from "react";
 import SidePanel from "@/components/layout/SidePanel/SidePanel";
 import { Button, Icon, ICONS, Tabs } from "@canonical/react-components";
 import Blocks from "@/components/layout/Blocks";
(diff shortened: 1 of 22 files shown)


src/features/mirrors/components/MirrorDetails/MirrorDetails.tsx
```diff
@@ -87,6 +87,19 @@ const MirrorDetails: FC = () => {
     },
   }));
 
+  useEffect(() => {
+    if (updateModal) {
+      openUpdateModal();
+    }
+  }, [openUpdateModal, updateModal]);
```
ethanashaw, line 94: I think this approach is not ideal, but it also seems necessary since the action is coming from a notification.

I wonder if we should move to a single global modal with an imperative action to make it pop up, kind of like the older side panel approach. It might make more sense with the way we use modal actions, especially since our `Modal` components have a lot of duplicate logic.

ethanashaw (approved):

src/features/mirrors/components/MirrorDetails/MirrorDetails.tsx
gesquivelgaghi: I think that makes sense actually. We can make a ticket for it

PR #654, opened 2026-06-03
Update Typescript 6 and other dependencies

ethanashaw:
- Updated most package dependencies, including TypeScript
- Modified `tsconfig.json` to fix TypeScript 6 deprecations
  - [`esModuleInterop: false` shouldn't be used anymore](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---esmoduleinterop-false-and---allowsyntheticdefaultimports-false)
  - [`moduleResolution` will be inferred as `"Bundler"`, which is correct for Vite](https://www.typescriptlang.org/tsconfig/#moduleResolution)
  - [Removed `baseUrl` and added the prefix to `paths` entry](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-6-0.html#deprecated---baseurl)
  - The Monaco editor import is fixed with a path interceptor


yurii-vasyliev (approved):
LGTM

I don't think we need the empty changeset file.

ethanashaw:
> I don't think we need the empty changeset file.

It's just for the checks, so that we can have that check as required

PR #656, opened 2026-06-04
fix: collapse table header actions and filters on medium screens [LNDENG-4486]

yurii-vasyliev:
## Summary

Jira: [LNDENG-4486](https://warthogs.atlassian.net/browse/LNDENG-4486)

On medium screens, instance-tab table headers overflowed: action buttons got cut off and crushed the search bar. This PR collapses header buttons into a contextual menu (via `ResponsiveButtons`) and filters into a "Filters" dropdown (via the existing `ResponsiveTableFilters`), keeping the search bar entirely visible.

### `ResponsiveButtons`: support `ConfirmationButton`s when collapsed

Previously a collapsed `ConfirmationButton` silently lost its confirmation modal (only `onClick` was carried over). The collapsed menu item now opens the same `ConfirmationModal`, hoisted outside the contextual menu so it survives the menu closing. Covered by new tests (confirm + cancel paths).

### Header refactors

| View | Buttons | collapseFrom |
|---|---|---|
| Instance → WSL | Install, Reinstall, Uninstall, Create new instance, Make compliant | `xl` |
| Instance → Activities (`ActivitiesActions`, also used on the Activities page) | Approve, Cancel, Redo | `xl` |
| Instance → Processes | End process, Kill process | `lg` |
| Instance → Security issues | Upgrade | — (single button, see below) |

> Following review, `ResponsiveButtons` no longer collapses a **lone** button into an "Actions" menu — it saves no horizontal space and just hides the action behind an extra click. So the Security issues header's single `Upgrade` action now stays inline at all widths. A caller that genuinely wants to collapse a single (e.g. long) action can still opt in via `alwaysCollapse`.

### Activities header

- Status/Type/Date range filters collapse into a "Filters" dropdown below `xl` (`ResponsiveTableFilters`, same as the Instances page); `ActivityTypeFilter` and `ActivitiesDateFilter` now forward the `inline` prop
- Fixed the search box being crushed to icon-width: broke the flexbox `min-width: auto` chain (with a 15rem floor) and made the actions group non-shrinking
- Added a gap between the filter group and the Actions menu

### Filter submenu positioning

The collapsed Filters menu sits at the right of the table header, so it and its nested submenus (e.g. `Filters → Status`) can run off the right edge. `ResponsiveTableFilters` measures the menu's position (on mount and on a passive `resize` listener) and drives **both** the dropdown alignment and the submenu direction from that single measurement:

- **Room on the right** (wide screens): the dropdown is left-aligned (extends right) and submenus open rightward with the default layout.
- **Near the right edge** (narrow screens): the dropdown flips to right-aligned (extends left) and the items reverse — chevron on the left, label on the right — opening leftward so they stay on screen.

The direction is settled before the menu is opened, so it never switches while interacting. Because `ResponsiveTableFilters`/`ResponsiveDropdownItem` are shared, this applies to both the Activities and the Instances headers.

### Packages header

- Added a gap between the Status filter and the Install button (the single filter stays inline, not collapsed)

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

[LNDENG-4486]: https://warthogs.atlassian.net/browse/LNDENG-4486?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


src/features/activities/components/ActivitiesHeader/ActivitiesHeader.tsx
gesquivelgaghi, line 1: the collapsed table filters open outside the screen
Image

src/features/activities/components/ActivitiesHeader/ActivitiesHeader.module.scss
```diff
@@ -9,13 +9,14 @@
...
 .actions {
+  column-gap: $sph--small;
   display: flex;
   flex-grow: 1;
+  flex-shrink: 0;
   justify-content: space-between;
+
+  > * {
+    flex-shrink: 0;
+  }
```
gesquivelgaghi, line 21: not sure if i'm missing something but these `flex-shrink`s dont actually seem to be doing anything

src/components/ui/ResponsiveButtons/ResponsiveButtons.tsx
gesquivelgaghi, line 1: the confirmation button needs to recognize the `has-icon` class 
Image

src/pages/dashboard/instances/[single]/tabs/security-issues/SecurityIssuesPanelHeader/SecurityIssuesPanelHeader.tsx
Comment (gesquivelgaghi) [inline src/pages/dashboard/instances/[single]/tabs/security-issues/SecurityIssuesPanelHeader/SecurityIssuesPanelHeader.tsx:1]: not sure if it really makes sense to collapse this button when its the only button in the header. It doesnt save any significant space, just makes the button one extra click away:

Image Image

src/features/activities/components/ActivitiesHeader/ActivitiesHeader.module.scss
yurii-vasyliev: They are load-bearing, just subtly. `.actions` is a flex item alongside the search box (`.wrapper`), which is the flexible one (`flex-shrink: 1`, floored at `min-width: 15rem`). Giving `.actions` `flex-shrink: 0` is what forces the search box to absorb the shrinking first instead of the actions/filters getting crushed — that's the "search box was squished to icon-width" fix from the PR description.

The inner `> * { flex-shrink: 0 }` then stops the filter group and the Actions menu from squishing against each other under tight widths (below the `xl` collapse point) — they keep their intrinsic size and the overflow is taken by the search box rather than the buttons.

So if you remove them the regression is "search box stays wide, action buttons get crushed" rather than the other way around. Happy to add a comment in the SCSS noting this if it helps readability.

src/components/ui/ResponsiveButtons/ResponsiveButtons.tsx
```diff
@@ -43,7 +47,15 @@ const ResponsiveButtons: FC<ResponsiveButtonGroupProps> = ({
...
-  const isLargeScreen = alwaysCollapse ? false : isLargerThanBreakpoint;
+  const [confirmationModalProps, setConfirmationModalProps] = useState<
+    ConfirmationButtonProps["confirmationModalProps"] | null
+  >(null);
+
+  // A lone button gains nothing from collapsing into an "Actions" menu — it
+  // saves no horizontal space and just hides the action behind an extra click.
+  const isLargeScreen = alwaysCollapse
+    ? false
+    : isLargerThanBreakpoint || buttons.length <= 1;
```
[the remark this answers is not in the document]
yurii-vasyliev: This is intentional, and follows reviewer feedback on this PR: collapsing the Security issues header's single `Upgrade` action into an "Actions" menu was flagged as not worthwhile — it saves no horizontal space and just puts the action behind an extra click. So a lone button now stays inline rather than collapsing at `collapseFrom`.

On your option (b): the escape hatch already exists — `alwaysCollapse` forces a single (e.g. long) action into the menu when a caller genuinely wants that, so a separate `collapseSingleButton` flag would be redundant. The only behaviour removed is *responsive* collapsing of a lone button below a breakpoint, which isn't used anywhere and is the exact case the reviewer asked us to drop.

I've updated the PR description (Header refactors table + a note) so it no longer implies the single Upgrade button collapses at `md`.

gesquivelgaghi (dismissed):
LGTM but you might wanna check with Dani if its fine to have the icons in different directions (or if they should all change when one changes) because he usually wants the text to be aligned in a list of buttons.

Image

Approving in case its fine, otherwise I'll re-review after changes

yurii-vasyliev:
@gesquivelgaghi , it's a bug actually, good catch! I've updated the code, please take another look.

src/components/filter/ResponsiveTableFilters/ResponsiveTableFilters.tsx
```diff
@@ -10,6 +11,10 @@ import classes from "./ResponsiveTableFilters.module.scss";
 import ResponsiveDropdownItem from "@/components/ui/ResponsiveDropdownItem";
 import type { FilterProps } from "@/components/filter/types";
 
+// Rough width a filter submenu needs to the right of the menu before it would
+// run off-screen; with less room than this we open submenus to the left.
+const SUBMENU_WIDTH_ESTIMATE_PX = 256;
```
gesquivelgaghi: estimating it doesnt seem like the best way to go about this (screenshot from Activities page). Is it possible to calculate the max width of the submenus and use that to determine whether to switch them all left? 

Or to be honest we could just always have them open left. It always has more space than the right, and it would certainly simplify things.

Image

gesquivelgaghi: Okay I actually just checked our pages in small screens and the ones that break are `Activities` and `Instances`. I'm not sure what min screen size we aim to support, but the main bottleneck is definitely the search/filter row, and the submenu issue (for both filters and page actions). These are the issues I found.

1. This filter submenu issue breaks in both Activities and Instances at under 520px width

Image
Image

<br />
<br />

2. The Instances search/filter row breaks at 480px width (potential fix could be moving the filter & column buttons to a new row under the search for small screens)

Image

<br />
<br />

3. The Instances page actions breaks at around 430px width

Image

src/features/activities/components/Activities/Activities.test.tsx
```diff
@@ -109,12 +110,15 @@ describe("Activities", () => {
...
+      setScreenSize("xxl");
       renderWithProviders(<Activities {...defaultProps} />);
 
       expect(screen.getByPlaceholderText("Search")).toBeInTheDocument();
       expect(
         screen.getByRole("button", { name: "Status" }),
       ).toBeInTheDocument();
+
+      resetScreenSize();
     });
```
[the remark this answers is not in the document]
yurii-vasyliev: Already handled — the only `setScreenSize` call is inside the `ActivitiesHeader integration` describe, which has an `afterEach(() => resetScreenSize())`, so the breakpoint mock is cleaned up even if an assertion fails first. No code change needed; resolving.

src/components/ui/ResponsiveButtons/ResponsiveButtons.tsx
```diff
@@ -204,6 +230,23 @@ const ResponsiveButtons: FC<ResponsiveButtonGroupProps> = ({
...
+            confirmationModalProps.close?.();
+          }}
+          onConfirm={(event) => {
+            setConfirmationModalProps(null);
+            confirmationModalProps.onConfirm?.(event);
+          }}
+        >
+          {confirmationModalProps.children}
+        </ConfirmationModal>
+      )}
```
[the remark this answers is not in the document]
yurii-vasyliev: Fixed in 6d2d8b9f8 — `onConfirm` now awaits the provided handler before clearing the modal state, so the confirm button's loading/disabled state stays meaningful during async confirms, and the modal stays open on error to allow a retry.

.changeset/tidy-keys-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": patch
+---
+
+Collapse table header action buttons into a menu on medium screens so the search bar stays visible in the WSL, Activities, Processes and Security issues views
```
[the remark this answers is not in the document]
yurii-vasyliev: Fixed in 6d2d8b9f8 — reworded the changeset: dropped the "Security issues collapses into a menu" claim (its lone Upgrade action no longer collapses) and added "and filters".

src/components/filter/ResponsiveTableFilters/ResponsiveTableFilters.tsx
yurii-vasyliev: Went with the "always open left" approach in 6d2d8b9f8 — removed the submenu-width estimation entirely; the Filters dropdown and its submenus now always open leftward. Simpler, and it fixes the off-screen cases on Activities/Instances. Thanks for the suggestion!

RyanLoi98 (approved):

RyanLoi98:
LGTM!

rubinaga (approved):
LGTM

PR #657, opened 2026-06-04
Add instance title to activities list [LNDENG-4522]

ethanashaw:
## Summary

The Activities table now loads instance names.

## Release Impact

Change type (tick one):

- [ ] Patch (fix)
- [x] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


yurii-vasyliev:
My main concern here is that every visible row with a `computer_id` now fires its own `GET computers/<id>`. One page of activities across N distinct instances = N parallel API calls per page, repeated on every pagination/filter change for uncached instances. React Query dedupes identical IDs and caches results of course, but we still have to call the endpoint for each row that returns a full Instance payload just to display the title.

In the SQL world we can definitely get the title from another table and include it in the main response. Ideally, such a component as `InstanceLink` shouldn't exist or should be used very rarely.

Also, if we have problems with the `GET computers` endpoint, on a page with 100 activities we'd trigger the global error toast 100 times.

src/features/instances/components/InstanceLink/InstanceLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+const InstanceLink: FC<InstanceLinkProps> = ({ instanceId }) => {
+  const { instance, isGettingInstance } = useGetInstance({ instanceId });
+
+  if (isGettingInstance) {
+    return <Spinner />;
+  }
+
+  if (!instance) {
+    return <NoData />;
```
yurii-vasyliev, line 20: According to the current flow, we'd show it as `ID: <id>` regardless of whether or not that instance exists. In this case we're not showing even that.

I'd consider to show something like this instead:
```
return (
      <Link to={ROUTES.instances.details.single(instanceId)}>
        ID: {instanceId}
      </Link>
    );
```

We can also add a warning/orange triangle icon on the left indicating that something is wrong with that instance.

src/features/instances/components/InstanceLink/InstanceLink.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+interface InstanceLinkProps {
+  readonly instanceId: number;
+}
+
+const InstanceLink: FC<InstanceLinkProps> = ({ instanceId }) => {
+  const { instance, isGettingInstance } = useGetInstance({ instanceId });
+
+  if (isGettingInstance) {
+    return <Spinner />;
```
yurii-vasyliev, line 16: I know that this is what we have it in some other places, but we could create a reusable skeleton component for such cases instead of lots of spinning elements in the table. Or fall back to `ID: <id>`, but that's less preferable.

yurii-vasyliev:
@ethanashaw , more clarity on this one. @Perfect5th said that including instance titles to the response was not there by default because these are different databases - for activities and the main one where instances live. However, we can add the same call to the main DB to get all the instance titles and include them in the response, so this PR will not be needed.

@ethanashaw or @rubinaga , feel free to create a ticket for the backend part of the task if you want to do it.

PR #658, opened 2026-06-08
Fix/scripts search empty state

RyanLoi98:
## Summary

Improve scripts empty-state behavior by showing a search-specific no-results message when filters return zero matches, while keeping the default empty state for unfiltered views; adds test coverage for both paths.

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


.changeset/whole-colts-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": minor
+---
+
+Improve scripts empty-state behavior by showing a search-specific no-results message when filters return zero matches, while keeping the default empty state for unfiltered views; adds test coverage for both paths.
```
gesquivelgaghi: This is more what the PR summary should be, not the changeset. 

The changeset message doesnt need to be so specific, we want it to be a very simple and clear description, since its for the users. They dont care about test coverage, for example.

```suggestion
Add a message for when scripts search returns no result 
```

src/features/scripts/components/ScriptsContainer/ScriptsContainer.tsx
```diff
@@ -14,6 +14,7 @@ const ScriptsContainer: FC = () => {
   const { currentPage, pageSize, search, status } = usePageParams();
 
   const { scripts, scriptsCount, isScriptsLoading } = useGetScripts();
+  const isFilteringScripts = !!search || !!status;
```
gesquivelgaghi: we can just get `search`  and `status` from `usePageParams` in `ScriptList`, we dont need to pass them from here. 

It works out better because we'll need to use `usePageParams` in `ScriptList` anyway when we update our sidepanels so there's no downside to just adding it straight there

src/features/scripts/components/ScriptsContainer/ScriptsContainer.test.tsx
```diff
@@ -31,17 +31,39 @@ describe("Scripts Empty State", () => {
 
   it("should show scripts when there are scripts with search", async () => {
     const searchText = scripts[0].title;
+    setEndpointStatus("default");
```
gesquivelgaghi: as a rule its usually better to keep these resets in a `beforeEach` instead of in each test

.gitignore
```diff
@@ -35,3 +35,4 @@ src/**/*.module.scss.d.ts
 /playwright-report/
 /playwright/
 .cert
+.prettierrc.json
```
gesquivelgaghi: why are we doing this?

src/features/scripts/components/ScriptList/ScriptList.tsx
```diff
@@ -177,10 +181,21 @@ const ScriptList: FC<ScriptListProps> = ({ scripts }) => {
...
+    isFeatureEnabled,
+    openViewPanel,
+  ]);
 
   return (
     <ResponsiveTable
+      emptyMsg={
+        isFilteringScripts
+          ? "No scripts found according to your search parameters."
+          : "No scripts found"
```
gesquivelgaghi: we dont need this branching because if there's no scripts and no search we wont show `ScriptList` at all, we just show `ScriptsEmptyState` instead. you can find this logic in `ScriptsContainer`

.gitignore
RyanLoi98: I noticed we didn't have a default prettier configuration, so I didn't want to accidentally commit mine to git.

src/features/scripts/components/ScriptsContainer/ScriptsContainer.test.tsx
RyanLoi98: Good advice, thanks!

.gitignore
yurii-vasyliev: We don't have it because we're using its defaults, but it doesn't mean we should add this config to gitignore, because we have a prettier job in the pipeline. You can just not commit yours in this PR, but I wouldn't add it to gitignore.

.changeset/whole-colts-thank.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": minor
```
yurii-vasyliev: I'd tag is a `patch` because it's rather a UX fix, not a new feature - minor inconsistency with how #644 (a similar "show a message when empty" change) was tagged `patch`.

yurii-vasyliev (changes requested):

src/features/scripts/components/ScriptList/ScriptList.tsx
```diff
@@ -177,10 +180,17 @@ const ScriptList: FC<ScriptListProps> = ({ scripts }) => {
...
+    handleExpand,
+    isFeatureEnabled,
+    openViewPanel,
+  ]);
 
   return (
     <ResponsiveTable
+      emptyMsg="No scripts found according to your search parameters."
       ref={getTableRowsRef}
       columns={columns}
```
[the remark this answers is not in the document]
gesquivelgaghi: instanceList only uses this conditional message because it doesnt have an emptyState, which is not what we want.

src/features/scripts/components/ScriptList/ScriptList.test.tsx
gesquivelgaghi: since you moved this to the `beforeEach` you dont need it here anymore. we only want to override the mock if it's meant to be different, like in the case of `vi.mocked(useAuth).mockReturnValue(authContextValuesWithoutFeatureFlag);`

src/features/scripts/components/ScriptsContainer/ScriptsContainer.test.tsx
```diff
@@ -3,24 +3,39 @@ import { expectLoadingState } from "@/tests/helpers";
...
-import { describe, expect, it } from "vitest";
+import { beforeEach, describe, expect, it } from "vitest";
 import ScriptsContainer from "./ScriptsContainer";
 
 describe("Scripts Empty State", () => {
+  beforeEach(() => {
+    setEndpointStatus("default");
+  });
+
   it("should empty state when there are no scripts", async () => {
```
gesquivelgaghi: unrelated to your changes, but could you fix this test title please?

```suggestion
  it("should show empty state when there are no scripts", async () => {
```

gesquivelgaghi (approved):

yurii-vasyliev (approved):
LGTM

PR #655, opened 2026-06-04
fix: move account name from API credentials to organization settings [LNDENG-3798]

yurii-vasyliev:
## Summary

Jira: [LNDENG-3798](https://warthogs.atlassian.net/browse/LNDENG-3798)

Our documentation uses the term "Account Name", so the API credentials view should not display it as "Name". Since the account name is an organization property (not a user property) and is not needed on the API credentials view, it has been moved to a more intuitive location:

- Removed the "Name" row (`account_name`) from the Organization table in **Account settings → API credentials**.
- Added a read-only "Account name" field to **Org. settings → General**, above "Organization's name", using the existing `ReadOnlyField` component. The account name is not changeable, and the tooltip explains that it is assigned at account creation and used to register clients with Landscape.
- Updated the affected tests.

## Release Impact

Pick the target branch (see [RELEASES.md](../RELEASES.md) for the full model):

- [ ] `dev` — internal testing → publishes to `ppa-build-dev`
- [x] `main` — next beta → publishes to `ppa-build`
- [ ] `release/YY.MM` — point release on a maintained cycle → publishes to `ppa-build-YY.MM` (and to `ppa-build-stable` if this is the currently-promoted branch). Specify cycle: `release/____`

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

> The change-type label is informational and only affects how the entry is rendered in the CHANGELOG. The actual version is computed from the branch and CalVer cycle — `pnpm changeset`'s `patch`/`minor` choice does not influence it.

## Checklist

- [x] **Changeset added** — I have run `pnpm changeset` (or `pnpm changeset --empty` if no CHANGELOG line is warranted) and committed the resulting `.md` file. Required for PRs targeting `main` and `release/*`; enforced by the `Changeset check` workflow.
- [ ] **UI verified** — I have verified the changes locally.
- [x] **Linting clean** — No linting errors are present (especially in `scripts/`).

[LNDENG-3798]: https://warthogs.atlassian.net/browse/LNDENG-3798?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #659, opened 2026-06-09
fix(ci): install dependencies before TICS analysis

yurii-vasyliev:
## Summary

The `TICS` job in `validate.yml` runs the TIOBE action without installing node dependencies. TICS uses **ESLint** as one of its analyzers, so with no `node_modules` it logs `The tool 'eslint' is not available…`, marks every changed TypeScript file as having *failed analysis*, and the **Coding Standard Violations Check** quality gate fails.

This has been the case since the workflow was restructured in #640 — every other job (`eslint`, `prettier`, `stylelint`, `unit-tests`, `e2e`) installs dependencies, but `tics-report` does not. It only surfaces on PRs that change `.ts`/`.tsx` files (e.g. #655).

Fix: add the same pnpm / Node / `pnpm install --frozen-lockfile` steps the other jobs use, before the TICS step, so ESLint is available for analysis.

## Release Impact

- [x] `main` — next beta

Change type:

- [x] Patch (fix)

## Checklist

- [x] **Changeset added** — empty changeset (CI-only change, no CHANGELOG entry warranted).
- [x] **UI verified** — n/a (CI workflow only).
- [x] **Linting clean** — workflow-only change.


rubinaga (approved):
LGTM

PR #661, opened 2026-06-10
refactor: disable switch channel button when no there are no available channels

rubinaga:
## Summary

The work in this PR is done to disable the button to switch channels on the snaps details sidepanel form. Before we were opening the switch form and disabling the user action which meant the form was useless. Now we are directly disabling the button so the user cannot click on it.

### Extra work
At the same time there were a lot of drive-bys done. The `EditSnap` component handled many different functionalities like `Hold`, `Unhold`, `Switch`, `Uninstall`. I removed it and made separate components for each of these. Furthermore refactored the api hooks from the `/hooks` folder to the `/api` folder with separate files per api request. 

Also changed the way selections were handled in `SnapsPanel`. It used to keep track of ids and installed snaps separately, instead I moved towards the approach used in `Activities.tsx` that uses `useSelection` to simplify things.

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


ethanashaw (dismissed):

src/features/snaps/components/HoldSnapForm/HoldSnapForm.tsx
```diff
@@ -0,0 +1,165 @@
...
+      );
+    }
+  };
+
+  return (
+    <Form onSubmit={formik.handleSubmit} noValidate>
+      {hasOneItem(installedSnaps) ? (
+        <p>
+          This will pause automatic updates for that particular snap.{" "}
+          {installedSnaps[0].snap.name} will maintain the current version.
```
RyanLoi98: Just nit picking (sorry), but I think the description could be a bit clearer, because this first paragraph only targets holding single snaps you can refer to its name directly. So I would suggest having this as the description:

```suggestion
          This will pause automatic updates for {installedSnaps[0].snap.name}, and it will maintain its current version.
```

src/features/snaps/components/UnholdSnapForm/UnholdSnapForm.tsx
```diff
@@ -0,0 +1,117 @@
...
+    },
+  });
+
+  return (
+    <Form onSubmit={formik.handleSubmit} noValidate>
+      {hasOneItem(installedSnaps) ? (
+        <p>
+          This will resume automatic updates for that particular snap.{" "}
+          {installedSnaps[0].snap.name} will be eligible for the latest version
+          upgrades based on the regular refresh schedule.
```
RyanLoi98: Again sorry to nit pick. But like before I would change this to be more clear. Since this only shows when you are unholding a single snap, we can refer to it directly. So I would change it to this:


```suggestion
          This will resume automatic updates for {installedSnaps[0].snap.name}, and it will be eligible for the latest version
          upgrades based on the regular refresh schedule.
```

RyanLoi98 (dismissed):
LGTM, I really like disabling the switch channel button and preventing the form from opening up rather than disabling the buttons in the form. It's much clearer.



Other than that I only had 2 minor suggestions how you could change descriptions to make things clearer.

src/features/snaps/components/HoldSnapForm/HoldSnapForm.tsx
rubinaga: thanks, added both

RyanLoi98 (approved):
LGTM!

ethanashaw (approved):

PR #663, opened 2026-06-10
Fix Ubuntu Pro mirror archive root [LNDENG-4627] [FEBUGS-287]

ethanashaw:
## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/src/features/mirrors/components/AddMirrorForm/AddMirrorForm.tsx
+++ b/src/features/mirrors/components/AddMirrorForm/AddMirrorForm.tsx
@@ -49,6 +49,23 @@ const AddMirrorForm: FC = () => {
   const isMirrorContentsLoading =
     ubuntuArchiveQuery.isPending || ubuntuEsmQuery.isPending;
 
+  const getArchiveRoot = (values: FormProps) => {
+    switch (values.sourceType) {
+      case "ubuntu-snapshots":
+        return `https://${UBUNTU_SNAPSHOTS_HOST}/ubuntu/${values.snapshotDate}`;
+
+      case "ubuntu-pro": {
+        const archiveRoot = new URL(values.proService);
+        archiveRoot.username = "bearer";
+        archiveRoot.password = values.token;
+        return archiveRoot.href;
+      }
+
+      default:
+        return values.sourceUrl;
+    }
+  };
+
   const formik = useFormik<FormProps>({
     initialValues: getInitialValues({
       ubuntuArchiveInfo,
@@ -86,13 +103,8 @@ const AddMirrorForm: FC = () => {
 
     onSubmit: async (values) => {
       try {
-        const archiveRoot =
-          values.sourceType === "ubuntu-snapshots"
-            ? `https://${UBUNTU_SNAPSHOTS_HOST}/ubuntu/${values.snapshotDate}`
-            : values.sourceUrl;
-
         const { data: newMirror } = await createMirror({
-          archiveRoot,
+          archiveRoot: getArchiveRoot(values),
           components: values.components.map((component) => component.trim()),
           displayName: values.name,
           architectures: values.architectures.map((architecture) =>
@@ -139,7 +151,7 @@ const AddMirrorForm: FC = () => {
 
   const proServiceOptions: SelectOption[] = ubuntuEsmInfo.map((proService) => ({
     label: proService.label,
-    value: proService.mirror_type,
+    value: proService.mirror_url,
     disabled: !isArchiveInfoValid(proService),
   }));
 
--- a/src/features/mirrors/components/SelectableMirrorContentsBlock/SelectableMirrorContentsBlock.tsx
+++ b/src/features/mirrors/components/SelectableMirrorContentsBlock/SelectableMirrorContentsBlock.tsx
@@ -32,7 +32,7 @@ const SelectableMirrorContentsBlock: FC<SelectableMirrorContentsBlockProps> = ({
 
       case "ubuntu-pro": {
         const proService = ubuntuEsmInfo.find(
-          ({ mirror_type }) => mirror_type === formik.values.proService,
+          ({ mirror_url }) => mirror_url === formik.values.proService,
         );
 
         return proService ? proService.distributions : [];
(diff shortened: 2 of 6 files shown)


src/features/mirrors/components/AddMirrorForm/AddMirrorForm.tsx
```diff
@@ -49,6 +49,23 @@ const AddMirrorForm: FC = () => {
...
+  const getArchiveRoot = (values: FormProps) => {
+    switch (values.sourceType) {
+      case "ubuntu-snapshots":
+        return `https://${UBUNTU_SNAPSHOTS_HOST}/ubuntu/${values.snapshotDate}`;
+
+      case "ubuntu-pro": {
+        const archiveRoot = new URL(values.proService);
+        archiveRoot.username = "bearer";
+        archiveRoot.password = values.token;
+        return archiveRoot.href;
```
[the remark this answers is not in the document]
ethanashaw: This needs to be fixed on the backend

yurii-vasyliev:
Nice — this also resolves the issue Copilot flagged on #665 (the token being collected but never sent).

One suggestion: the existing Ubuntu Pro test in `AddMirrorForm.test.tsx` ends with

```ts
expect(mockCreateMirror).toHaveBeenCalledExactlyOnceWith(
  expect.objectContaining({}),
);
```

which passes regardless of payload, so the new `getArchiveRoot` behavior isn't actually pinned by any test. Tightening it would protect the fix from regressing:

```ts
expect(mockCreateMirror).toHaveBeenCalledExactlyOnceWith(
  expect.objectContaining({
    archiveRoot: `https://bearer:${token}@esm.ubuntu.com/infra/ubuntu/`,
  }),
);
```

(adjust the path to whichever Pro service the mocks preselect — and it'd also catch accidental double-encoding of pre-encoded tokens in the URL password setter.)

gesquivelgaghi (dismissed):
LGTM

src/features/mirrors/components/SelectableMirrorContentsBlock/SelectableMirrorContentsBlock.tsx
```diff
@@ -32,7 +32,7 @@ const SelectableMirrorContentsBlock: FC<SelectableMirrorContentsBlockProps> = ({
 
       case "ubuntu-pro": {
         const proService = ubuntuEsmInfo.find(
-          ({ mirror_type }) => mirror_type === formik.values.proService,
+          ({ mirror_url }) => mirror_url === formik.values.proService,
```
yurii-vasyliev, line 35: This lookup now matches on `mirror_url`, but `SelectableMirrorContentsBlock.test.tsx:79` still seeds `proService: proService.mirror_type`. Since `mirror_type` (e.g. `one-of-each`) never equals a `mirror_url` (`https://esm.ubuntu.com/one-of-each/ubuntu/`), `find` returns `undefined` → `distributions = []`, so the *"renders readonly components when there is only one option"* test actually renders 3 editable comboboxes, not readonly fields. I verified this by running the test on this branch: seeded with `mirror_type` it renders 3 comboboxes; seeded with `mirror_url` it renders 0. It only stays green because the test has no assertions.

Suggest updating the test to `proService.mirror_url` and adding an assertion so it can't silently rot again, e.g. `expect(screen.queryAllByRole("combobox")).toHaveLength(0)`.

src/features/mirrors/components/AddMirrorForm/AddMirrorForm.tsx
```diff
@@ -49,6 +49,23 @@ const AddMirrorForm: FC = () => {
...
 
+  const getArchiveRoot = (values: FormProps) => {
+    switch (values.sourceType) {
+      case "ubuntu-snapshots":
+        return `https://${UBUNTU_SNAPSHOTS_HOST}/ubuntu/${values.snapshotDate}`;
+
+      case "ubuntu-pro": {
+        const archiveRoot = new URL(values.proService);
+        archiveRoot.username = "bearer";
+        archiveRoot.password = values.token;
```
yurii-vasyliev, line 60: The token is now embedded in `archiveRoot` as `https://bearer:<token>@…`. `MirrorDetails.tsx` and `EditMirrorForm.tsx` render `mirror.archiveRoot` verbatim (as visible text and as an `href`), so if the API echoes the credentials back on read, the Pro token would be displayed in the UI. There's already a `getStrippedUrl()` helper in `AddMirrorForm/helpers.ts` that strips `…@` credentials, but it's currently unused — it looks intended for exactly this display path. Worth confirming what the backend returns for a Pro mirror's `archiveRoot` and wiring up credential-stripping for display if it includes the token.

yurii-vasyliev (commented):
Production fix looks correct (the `mirror_type` → `mirror_url` switch is consistent across the three call sites, and `getSourceType` still resolves to "Ubuntu Pro" since `UBUNTU_PRO_HOST` is the host of every `mirror_url`). Two things worth addressing — details inline.

src/features/mirrors/components/AddMirrorForm/AddMirrorForm.tsx
ethanashaw: There is a ticket to fix this on the backend side: https://warthogs.atlassian.net/browse/LNDENG-4632

yurii-vasyliev: Non-blocking — opened #667 as a follow-up that wires up the (currently dead) `getStrippedUrl` helper to strip the `bearer:<token>@` credentials before displaying the archive root in mirror details and the edit form. It's a no-op until this PR merges (nothing embeds the token before then) and a no-op for non-Pro mirrors, so it doesn't need to gate this PR. Worth confirming separately whether the API echoes the token back in `archiveRoot` on read.

yurii-vasyliev (approved):
LGTM

gesquivelgaghi (approved):

PR #673, opened 2026-06-18
feat: gate the instance report view behind the instance-reports feature flag

yurii-vasyliev:
## Summary

> [!NOTE]
> **Scope reduced after #685.** The report side panel redesign that this PR originally carried was absorbed into #685 (which built on this branch) and is already on `main`. After merging `main` back in, this PR now only delivers the remaining piece: the feature-flag gating.

Gates the instances report view behind the server-driven `instance-reports` feature flag (from the `features/` endpoint), replacing the build-time `VITE_REPORT_VIEW_ENABLED` env var:

- "View report" in the Operations menu and the `sidePath=report` side panel are shown only when `isFeatureEnabled("instance-reports")` is true.
- Removes `REPORT_VIEW_ENABLED` from `constants.ts`, its `VITE_*` typing, `.env.production` / `.env.local.example` entries, and the now-unused workflow env vars.
- Tests assert the gating against a mocked `features/` endpoint (`enabled: false` override) instead of spying on the constant.

Requires the server side: canonical/landscape-server#1476 (adds the feature flag; on by default).

## Release Impact

Change type:

- [x] Minor (feature)

## Checklist

- [x] **Changeset added** — `.changeset/reports-side-panel.md` (minor).
- [x] **UI verified** — verified locally.
- [x] **Linting clean** — `tsc`, eslint, and the touched test suites pass.

## Testing against MSW (no server required)

The mocked `features/` endpoint returns `instance-reports` as **enabled** (kept present even when the endpoint status is mocked as empty), so "View report" is available against MSW by default. To test the gated state, override the `features/` endpoint with `server.use(...)` returning the feature as disabled.


src/features/reports/types.ts
```diff
@@ -0,0 +1,17 @@
+export interface ReportBucket {
+  count: number;
+  computer_ids?: number[];
+}
```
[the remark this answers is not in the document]
yurii-vasyliev: Done in a0a4e2f7f — made `computer_ids` required on `ReportBucket` (the V2 endpoint always returns ids per bucket, empty array when count is 0) and removed the now-redundant `?? []` fallbacks in `ReportView`.

src/features/instances/selectedInstancesStore.ts
```diff
@@ -0,0 +1,27 @@
...
+
+let selectedInstanceIds: number[] = [];
+const listeners = new Set<() => void>();
+
+export const setSelectedInstanceIds = (ids: number[]): void => {
+  selectedInstanceIds = ids;
+  listeners.forEach((listener) => {
+    listener();
+  });
+};
```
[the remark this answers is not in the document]
yurii-vasyliev: Done — `setSelectedInstanceIds` now stores `Object.freeze([...ids])`, so callers can't mutate the snapshot in place and desync the `useSyncExternalStore` listeners.

src/features/reports/components/MetricBarTable/MetricBarTable.tsx
```diff
@@ -0,0 +1,67 @@
...
+  });
+
+  return (
+    <ReportLegendTable
+      variant="bar"
+      markerHeader="Instances"
+      labelHeader={labelHeader}
+      countHeader={countHeader}
+      rows={legendRows}
+    />
```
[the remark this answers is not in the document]
yurii-vasyliev: Done — relabeled the bar column header to **Proportion** so it no longer duplicates the "Instances" count header (the bar is the share, the right column is the count).

src/features/instances/components/InstancesPageActions/InstancesPageActions.test.tsx
```diff
@@ -126,7 +124,7 @@ describe("InstancesPageActions", () => {
     });
 
     it("'View report' menu item should not be visible when feature disabled", async () => {
-      vi.spyOn(Constants, "REPORT_VIEW_ENABLED", "get").mockReturnValue(false);
+      setEndpointStatus({ status: "empty", path: "features" });
```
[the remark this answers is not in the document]
yurii-vasyliev: Done — the gating test now overrides the `features/` endpoint via `server.use(...)` to return `instance-reports` with `enabled: false`, which matches the realistic server response rather than an empty list.

src/features/reports/components/ReportLegendTable/ReportLegendTable.module.scss
```diff
@@ -0,0 +1,87 @@
...
+
+th:first-child,
+td:first-child {
+  padding-left: 0;
+}
+
+th:last-child,
+td:last-child {
+  padding-right: 0;
+}
```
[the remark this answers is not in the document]
yurii-vasyliev: Good catch — nested the `th`/`td` element rules (and the first/last-child rules) under the local `.table` class so they no longer compile to global selectors and can't leak table styling app-wide.

src/tests/server/handlers/reports.ts
```diff
@@ -1,54 +1,52 @@
...
+  http.get(`${API_URL}computers/report`, () => {
+    if (shouldApplyEndpointStatus("computers/report")) {
+      const { status } = getEndpointStatus();
 
-      if (endpointStatus.status === "empty") {
-        return HttpResponse.json("");
+      if (status === "error") {
+        throw createEndpointStatusNetworkError();
       }
     }
```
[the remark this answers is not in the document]
yurii-vasyliev: Done — the handler now returns a zeroed `emptyComplianceReport` (total 0, empty buckets) for the `empty` status, so the endpoint-status controller is consistent for this endpoint and devs can simulate the empty report.

src/features/instances/selectedInstancesStore.ts
```diff
@@ -0,0 +1,29 @@
...
+  // Copy and freeze so callers can't mutate the store's snapshot in place
+  // (which would change state without notifying useSyncExternalStore listeners).
+  selectedInstanceIds = Object.freeze([...ids]) as number[];
+  listeners.forEach((listener) => {
+    listener();
+  });
+};
+
+export const getSelectedInstanceIds = (): number[] => selectedInstanceIds;
+
```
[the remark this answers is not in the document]
yurii-vasyliev: Done — `getSelectedInstanceIds`/`setSelectedInstanceIds` now use `readonly number[]`, which removed the `as number[]` cast and made accidental mutation a compile-time error. Propagated `readonly` through `ReportView` and `ReportForm`.

src/features/reports/components/ReportView/ReportView.tsx
```diff
@@ -0,0 +1,276 @@
...
+  // opened (or last regenerated); the live selection is only observed to
+  // tell the user when the snapshot has gone stale.
+  const [reportIds, setReportIds] = useState(instanceIds);
+  const currentIds = useSyncExternalStore(
+    subscribeToSelectedInstanceIds,
+    getSelectedInstanceIds,
+  );
+  const selectionChanged =
+    currentIds.length !== reportIds.length ||
+    !currentIds.every((id) => reportIds.includes(id));
```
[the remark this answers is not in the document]
yurii-vasyliev: Done — `reportIds` is now wrapped in a `Set` for the membership check, so `selectionChanged` is O(n) instead of O(n²) on large multi-selections.

src/features/reports/types.ts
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — `computer_ids` is required (`number[]`) on this type, so segment counts and deep links are always populated. Resolving.

src/features/instances/selectedInstancesStore.ts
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — `setSelectedInstanceIds` stores `Object.freeze([...ids])`, so callers can't mutate the store's snapshot in place. Resolving.

src/features/reports/components/MetricBarTable/MetricBarTable.tsx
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — the proportional-bar column now uses a distinct `markerHeader="Proportion"`, separate from the count column header. Resolving.

src/features/instances/components/InstancesPageActions/InstancesPageActions.test.tsx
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — the test now returns the `/features` list with `instance-reports` set to `enabled: false` (rather than an empty list), so it genuinely exercises the disabled-feature path. Resolving.

src/features/reports/components/ReportLegendTable/ReportLegendTable.module.scss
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — `th`/`td` are nested under the local `.table` class (with a comment noting CSS modules don't scope bare element selectors), so they no longer compile to global rules. Resolving.

src/tests/server/handlers/reports.ts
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — the handler now returns `emptyComplianceReport` when the endpoint status is `empty`, so that scenario can be simulated. Resolving.

src/features/instances/selectedInstancesStore.ts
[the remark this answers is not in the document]
yurii-vasyliev: Addressed — the store now types `selectedInstanceIds` as `readonly number[]` (field, setter and getter), so accidental mutation is a compile-time error. Resolving.

src/features/instances/selectedInstancesStore.ts
```diff
@@ -0,0 +1,31 @@
+/**
```
rubinaga, line 1: I don't think we should introduce this pattern. The side panel shouldn't be trying to track a moving selection at all. Once the panel is open it should work with the snapshot it was given. That would let us remove this store entirely, and also remove the "Selection has changed" notification that was added alongside it.

src/features/reports/components/ReportForm/ReportForm.tsx
```diff
@@ -30,28 +32,41 @@ const downloadCSV = (csvString: string, filename: string) => {
...
+        .required("This field is required"),
       reportByCve: Yup.boolean(),
     }),
-    // todo: figure out how to use form values to submit
-    onSubmit: () => {
-      downloadCSV(getCsvComplianceDataResult?.data || "", "report.csv");
+    onSubmit: async () => {
+      // `refetch` is declared just below: the query must read the latest form
+      // values, so the hook is set up after the form it depends on.
+      // eslint-disable-next-line no-use-before-define
```
rubinaga: Can we remove the eslint suppressor here

src/pages/dashboard/instances/InstancesPage/InstancesPage.tsx
```diff
@@ -26,6 +31,21 @@ const InstancesPage: FC = () => {
 
   const [selectedInstances, setSelectedInstances] = useState<Instance[]>([]);
 
+  // Clear the selection when the search query changes (e.g. following a report
+  // deep link), matching how header filter changes reset it. Resetting state
+  // during render (React's documented pattern) avoids an extra commit.
+  const [trackedQuery, setTrackedQuery] = useState(query);
```
rubinaga: Could useSelection be used here instead? It already auto-clears the selection when the items list changes, which might make this trackedQuery workaround unnecessary.

src/features/reports/components/ReportView/ReportView.tsx
```diff
@@ -0,0 +1,275 @@
...
+  const { total } = report;
+
+  // Every instance lands in exactly one bucket, worst first: instances with
+  // USNs outstanding for over 60 days take priority, the rest are split by the
+  // longest time they took to apply a USN (the fixed_in buckets are
+  // cumulative). Whatever the server cannot place in a named interval — USNs
+  // still unpatched for under 60 days, or patched after more than 60 days — is
+  // reported honestly as "Other". We carry the exact computer ids per bucket so
+  // the deep link matches the report's count precisely.
+  const pendingIds = report.usn_pending_over_60_days.computer_ids;
```
rubinaga: Can this bucket derivation logic live in a separate hook or a helpers.ts file? I think it's complex enough to be moved somewhere else,. Can also be tested in isolation easier I think

src/features/reports/components/ReportView/ReportView.tsx
```diff
@@ -0,0 +1,275 @@
...
+          total={total}
+          segments={patchSegments}
+          labelHeader="Time to patch USNs"
+          countHeader="Instances"
+        />
+      </section>
+      <p className="u-text--muted p-text--small">
+        {`Report generated ${moment(report.generated_at).format(DISPLAY_DATE_TIME_FORMAT)}`}
+      </p>
+      <SidePanelFormButtons
```
rubinaga: I think it was mentioned in a design meeting that the convention is that action buttons belong at the top when the panel is informational, so I guess this can be moved to the top in this sidepanel.

src/features/reports/components/ReportLegendTable/ReportLegendTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+const ReportLegendTable: FC<ReportLegendTableProps> = ({
+  variant,
+  markerHeader,
+  labelHeader,
+  countHeader,
+  rows,
+  activeKey,
+  onRowHover,
+}) => (
+  <table
```
rubinaga: Can't we use a <ModularTable /> component here instead of doing the table manually

src/features/reports/components/ReportView/ReportView.tsx
```diff
@@ -0,0 +1,275 @@
...
+    setReportIds(currentIds);
+    setSidePanelTitle(
+      `Report for ${pluralize(currentIds.length, ["instance"], "exact")}`,
+    );
+  };
+
+  // A shareable deep link to the instances list filtered to exactly these
+  // instances. We filter by id (not by a semantic term) so the result matches
+  // the report's selection-scoped count exactly.
+  const instancesHref = (ids: readonly number[]): string | undefined =>
```
rubinaga, line 99: I'm not sure this will scale too well, especially when the number of instances grows. It would put a very long `ID:1 OR ...` in the searchbox, and the url as well. Their initial selection could be large, or we don't plan to support selecting the whole fleet for this operation?

src/features/reports/components/ReportView/ReportView.tsx
```diff
@@ -0,0 +1,275 @@
...
+        </p>
+        <ReportDonutChart
+          total={total}
+          segments={patchSegments}
+          labelHeader="Time to patch USNs"
+          countHeader="Instances"
+        />
+      </section>
+      <p className="u-text--muted p-text--small">
+        {`Report generated ${moment(report.generated_at).format(DISPLAY_DATE_TIME_FORMAT)}`}
```
rubinaga: Is the time the report generated always "right now", or does it change in the backend (cached or something like that)? If it's always right now, I don't see the benefit of this

src/features/reports/components/ReportView/ReportView.tsx
```diff
@@ -0,0 +1,275 @@
...
+      count: report.covered_by_upgrade_profiles.count,
+      total,
+      color: "link",
+      countHref: instancesHref(report.covered_by_upgrade_profiles.computer_ids),
+      onCountActivate: closeSidePanel,
+      countAriaLabel: `View the ${report.covered_by_upgrade_profiles.count} instances covered by upgrade profiles`,
+    },
+    {
+      key: "contacted",
+      label: "Contacted in last 5 min",
```
rubinaga, line 261: Nit: do you think this text is good enough as is?

rubinaga (changes requested):

src/features/reports/components/ReportForm/ReportForm.tsx
yurii-vasyliev: Done — removed the suppressor. `refetch` is now bridged through a ref that's synced in a `useEffect`, so the submit handler calls the latest refetch without the forward reference.

src/features/reports/components/ReportView/ReportView.tsx
yurii-vasyliev: Done — moved the derivation into `ReportView/helpers.ts` as a pure `derivePatchBuckets(report)`, now covered by isolated unit tests in `helpers.test.ts`.

src/features/reports/components/ReportView/ReportView.tsx
yurii-vasyliev: Done — moved the Download-as-CSV action to the top of the panel (right-aligned) and dropped the sticky footer, per the informational-panel convention.

src/features/reports/components/ReportLegendTable/ReportLegendTable.tsx
yurii-vasyliev: Done — rebuilt the legend on `ModularTable`. The donut hover-sync and row dimming go through `getRowProps`, and the markers, tooltips and deep-link counts through column `Cell`s, so both the donut and status-bar variants keep their behaviour.

src/features/instances/selectedInstancesStore.ts
yurii-vasyliev: This is intentional for now: we deliberately observe the live selection so the panel can surface the "Selection has changed" notice and offer a regenerate. I'd like to keep it until the design review decides whether the panel should work purely from an open-time snapshot — happy to drop the store (and the notice) then.

src/pages/dashboard/instances/InstancesPage/InstancesPage.tsx
yurii-vasyliev: This `trackedQuery` is tied to the intentional selection-tracking above (`selectedInstancesStore`), so I'd like to revisit switching to `useSelection` together with that store as part of the design review rather than in isolation.

src/features/reports/components/ReportView/ReportView.tsx
yurii-vasyliev: We don't support whole-fleet selection for this action — the selection is capped at 100 instances, so the `id:1 OR ...` query (and the URL) stays bounded. If we ever raise that cap we can move to a server-side selection token, but it isn't a concern at the current limit.

src/features/reports/components/ReportView/ReportView.tsx
yurii-vasyliev: It's the backend's `generated_at`, not the client's "now" — the report can be served from cache, so this tells the user how fresh the data is. Keeping it for that reason.

src/features/reports/components/ReportView/ReportView.tsx
yurii-vasyliev: I think it reads fine as-is, so I'll leave it — happy to tweak if you have specific wording in mind.

ethanashaw:
@yurii-vasyliev, looks like there is a minor conflict if you could fix that before we review

marqode:
@yurii-vasyliev I think [this PR](https://github.com/canonical/landscape-ui/pull/685/changes) is the most up to date for the Reports feature, since I've been making design changes there. If you want I could split that PR into Reports (replacing this one) and Report Exports so that they're easier to review. Just thinking if we want to merge that branch and this one eventually there will be a lot of conflicts to resolve.

src/features/instances/components/InstancesPageActions/InstancesPageActions.tsx
```diff
@@ -23,7 +22,7 @@ const Upgrades = lazy(
   async () => import("@/features/upgrades/components/Upgrades"),
 );
 const ReportView = lazy(
-  async () => import("@/pages/dashboard/instances/ReportView"),
+  async () => import("@/features/reports/components/ReportView"),
 );
```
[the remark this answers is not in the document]
yurii-vasyliev: Keeping the deep import here, for consistency: every other lazy-loaded panel in this file already deep-imports the component module (`RunInstanceScriptForm`, `Upgrades`, `AttachTokenForm`, `ReplaceTokenForm`, ...), because `React.lazy` needs a default export while the feature barrels re-export named. Mapping the barrel's named export to a default just for this one import would make it the odd one out. If we want barrel-only imports for lazy components, that's a codebase-wide convention change better done separately.

yurii-vasyliev:
Updated from the latest `main` and resolved the conflicts.

The conflict was with #685, which had built on this branch and already merged the report panel redesign (with @marqode's design changes and the async TSV export flow) into `main`. I resolved by taking `main`'s implementation for all of the overlapping report/selection code, so this PR now only carries the piece that #685 did not: **gating the report view behind the server-driven `instance-reports` feature flag** instead of the build-time `VITE_REPORT_VIEW_ENABLED` env var (per the team decision to hide the feature behind a flag). Title, description and changeset updated accordingly — the diff is now ~11 files.

On the open review threads: the comments on `selectedInstancesStore`, `trackedQuery`/`useSelection`, `ReportForm` error handling, the legend table and the ReportView aria-labels all concern code that is no longer part of this PR's diff — those files now come verbatim from `main` (#685), which reworked or removed the code in question (e.g. `ReportForm` was replaced by `ReportExportForm`, and the `trackedQuery` selection-clearing was dropped). Any remaining improvements there would be follow-ups against `main`.

marqode (approved):
LGTM!

marqode (dismissed):
Nit: Summary still has "Requires the server side: https://github.com/canonical/landscape-server/pull/1476 (adds the feature flag; on by default)." but that PR is closed. Also I thought feature flags were off by default?

rubinaga (commented):
For TSV exports, the backend suggested we handle disabling on the UI side since in self-hosted users can enable feature flags to test things out. Is this okay with a feature flag or should it be controlled on our side if its enabled or not.

yurii-vasyliev:
Merged \`main\` (\`f7373fb\`) back in and resolved the conflict — it was import-only in \`InstancesPageActions.test.tsx\`, where this branch adds \`server\` / \`generatePaginatedResponse\` (for the \`features/\` override) and \`main\` added \`waitFor\` (for the \#766 remove-success assertion). Kept both.

All checks are green: ESLint, Prettier, Stylelint, Vitest, Changeset, Playwright (SaaS + self-hosted).

@rubinaga — your review is still marked as changes-requested, but it predates the scope reduction. Every comment you left was on \`ReportView.tsx\`, \`ReportLegendTable.tsx\` and \`selectedInstancesStore.ts\`, all of which were absorbed into \#685 and are already on \`main\`; each was addressed in the 2026-07-03 round (buttons moved to the top of the informational panel, legend rebuilt on \`ModularTable\`, selection query bounded by the 100-instance cap, \`generated_at\` is the backend value rather than client "now").

The diff is now just the feature-flag gating — 11 files, none of them under \`src/features/reports/\`. Could you take another look and dismiss/re-review when you get a chance?

rubinaga (dismissed):
LGTM

marqode (dismissed):
looks like it needs a fresh merge main but otherwise LGTM!

marqode (approved):
LGTM!

rubinaga (approved):
LGTM

PR #674, opened 2026-06-18
feat: release-process overhaul — main as CHANGELOG baseline + YY.MM.N point releases

yurii-vasyliev:
## What

Overhaul of the release/versioning pipeline:

1. **`main` is no longer a publish target** — it's the integration trunk + CHANGELOG baseline. Its Changeset versioning moves to a dedicated `changeset-version.yml` (no build, no tag); merging its Version Packages PR stays optional and just drains the changeset backlog.
2. **Versioning reworked** (`scripts/calculate-version.cjs`) so the CalVer point segment is used properly:
   - `release/YY.MM` → `YY.MM.0.<count>` (cycle base)
   - `point/YY.MM.N` → `YY.MM.N.<count>` (N ≥ 1; `point/YY.MM.0` is rejected — it would collide with the cycle base)
   - `-rc` suffix until the branch is listed in the new `RELEASED_BRANCHES` Actions variable, applied **uniformly** to `release/*` and `point/*` (candidate → released promotion)
   - cycle + point number pinned by branch name (no calendar drift on rebuild)
   - `<count>` is **tag-anchored**: release-worthy commits since the `.0` build tag, excluding `chore`/`ci`/`docs`/`test`/`style` so housekeeping commits don't burn a version
3. **Per-release PPA targets** (`ppa-build-YY.MM.N`) so concurrent point releases don't collide on a shared `ppa-build-point`.
4. **Build pipeline** (`release-and-build.yml`):
   - tags the source commit `v<version>` after a successful deploy — an immutable version↔commit record and the next build's counter anchor
   - skips build/publish/tag when `v<version>` is already tagged, so housekeeping-only pushes are no-ops and every tag stays unique
   - `workflow_dispatch` to (re)build a branch after promoting it via the variable, guarded to `release/*` / `point/*` refs only
5. **`dev` removed** from the pipeline (script, workflows, docs) — the branch no longer exists. `run-tests.yml` (dead `dev`-only trigger, already covered by `validate.yml`) is deleted; the unrelated `environment: dev` (a GitHub Environment for test secrets) is left in place.

## Notes

- `RELEASED_BRANCHES` governs both tiers (`release/*` and `point/*`) — listing a branch is what drops its `-rc`. It's already set in repo settings.
- The `point/*` branch migration to the new `point/YY.MM.N` names is handled operationally (separate from this PR).
- Direct-push backport docs in §4/§5 are rewritten to the PR-based flow the branch protection now requires.
- See `RELEASES.md` for the full operator guide.


RyanLoi98 (dismissed):
LGTM, just consider copilot's suggestions about changing the commenting.

yurii-vasyliev:
Folded in the improvements from the now-closed #684 (commit fd6d370d1):

- **Tag-anchored counter** in `calculate-version.cjs` replacing `origin/main..HEAD`, excluding `chore/ci/docs/test/style` so housekeeping commits don't burn a version.
- **Uniform `-rc`**: the candidate rule now applies to `release/*` (the cycle base) as well as `point/*`, both governed by `RELEASED_POINT_BRANCHES` (now also accepts `refs/heads/*` form). Listing a branch — release or point — is what drops `-rc`. Note this means the stable cycle base ships `-rc` until its `release/YY.MM` branch is listed.
- **Skip-if-tagged**: housekeeping-only pushes are no-ops; install/changesets/build steps gated behind `should_build`.
- **Source tagging** (`v<version>`) after deploy — fills the gap §7 called out, and doubles as the next build's anchor.
- **`workflow_dispatch`** to (re)build a branch after promoting it via the variable.
- `RELEASES.md` updated to match all of the above.

Dry-run: `release/26.10`→`26.10.0.0-rc`, `point/26.10.1` (listed)→`26.10.1.0`, `point/26.10.2`→`26.10.2.0-rc`, listed `release/26.10`→`26.10.0.0`.

scripts/calculate-version.cjs
```diff
@@ -1,34 +1,41 @@
...
+ * formatting or CI-only commit does not burn a version number. When no `.0`
+ * tag exists yet — a freshly cut branch, or the first build under this scheme —
+ * the current HEAD *is* the cut: n = 0, and CI tags it `.0`. Every later build
+ * counts up from that anchor, independent of main and immune to main drift.
  */
 
+// Conventional-commit types that do not advance the version on release/point
+// branches. A `fix(ci):` still counts (its type is `fix`); only a `chore(ci):`
+// — type `chore` — is skipped.
+const SKIP_TYPES = /^(chore|ci|docs|test|style)(\([^)]*\))?!?:/i;
```
RyanLoi98, line 37: Just wanted to bring up a potential edge case, but an incorrect commit message could also cause an accidental bump of the version numbering (eg: "clarified prompt message" as a commit message would cause a version bump). Not sure how we could stop this but just wanted to mention it.

.github/workflows/release-and-build.yml
```diff
@@ -81,10 +100,27 @@ jobs:
           fi
           echo "publish_stable=$PUBLISH_STABLE" >> $GITHUB_OUTPUT
 
+          # Skip the build/publish/tag entirely when this exact version was
+          # already built and tagged. A push that adds only skipped commit types
+          # (chore/ci/docs/test/style) leaves the counter unchanged, so the
+          # version matches the last build -> its tag exists -> no-op. This also
+          # keeps every v<version> tag unique.
```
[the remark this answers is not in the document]
RyanLoi98: If I understand correctly, this means that any commit that does not advance the computed version will be skipped by the publish flow. 

This could be an issue if we needed to republish without a release-worthy commit. Because the current skip filter (chore|ci|docs|test|style) means housekeeping-only changes won’t produce a new version, so they won’t trigger a rebuild/publish. So we will need need to keep this in mind that if we have changes that are encompassed by the skip filter, that way we can also make a commit that does trigger a new version, so the changes can be published.

Not a big deal since we could always make a change to bump the version.

RyanLoi98 (dismissed):
LGTM.

The only nit picky things I wanted to mention were my comments about the accidental version bumping with an incorrectly made commit message, and the inability for previously published branches to be republished without changes that cause a version bump (which honestly isn't a big deal since we could easily just make a change with a version bump).


Also I would look at copilot's suggestions about fixing some comments and descriptions.

scripts/calculate-version.cjs
yurii-vasyliev: You're right: the filter keys off the conventional-commit type prefix (chore/ci/docs/test/style), not the actual content — so an un-prefixed message like clarified prompt message is treated as release-worthy and bumps the count. But it's better to at least filter out such things because now we bump the count on every single commit. Plus, we kind of use the "conventional commits" methodology anyway.

ethanashaw (approved):

scripts/calculate-version.cjs
RyanLoi98: Yup I agree!

RyanLoi98 (approved):
LGTM!

PR #675, opened 2026-06-18
feat: add exports for instances and activities

rubinaga:
## Summary

Added exports feature for instances and activities (including activities in a specific instance)

## Release Impact

Change type (tick one):

- [ ] Patch (fix)
- [x] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


src/features/instances/api/useExportInstancesTsv.ts
```diff
@@ -0,0 +1,36 @@
...
+    AxiosError<ApiError>,
+    CreateInstancesExportJobParams
+  >({
+    mutationFn: async (params) =>
+      authFetch.post<InstancesExportJob>("computers/exports", params),
+    onSuccess: async () => {
+      await queryClient.invalidateQueries({
+        queryKey: ["instances-export-jobs"],
+      });
+    },
```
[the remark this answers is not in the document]
rubinaga: fixed

src/features/activities/api/useExportActivitiesTsv.ts
```diff
@@ -0,0 +1,37 @@
...
+    AxiosError<ApiError>,
+    CreateActivitiesExportJobParams
+  >({
+    mutationFn: async (params) =>
+      authFetch.post<ActivitiesExportJob>("activities/exports", params),
+    onSuccess: async () => {
+      await queryClient.invalidateQueries({
+        queryKey: ["activities-export-jobs"],
+      });
+    },
```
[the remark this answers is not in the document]
rubinaga: fixed

src/features/exports/components/ExportDetailsSidePanel/ExportDetailsSidePanel.tsx
```diff
@@ -0,0 +1,208 @@
...
+    }
+
+    try {
+      const result = await onDownload(job);
+      if (result) {
+        notify.success({
+          title: "TSV download started",
+          message: `${job.name} has been downloaded and removed from the export list.`,
+        });
+      }
```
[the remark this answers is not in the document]
rubinaga: on download, the export is removed

src/features/activities/components/ActivitiesExportForm/ActivitiesExportForm.tsx
```diff
@@ -0,0 +1,291 @@
...
+          selected_field_ids: fieldsToExport.map((field) => field.id),
+          retain_until: moment(values.retainUntil).toISOString(),
+        });
+        const job = response.data;
+
+        closeSidePanel();
+        notify.success({
+          title: "TSV export in progress",
+          message: `Your activities export "${values.name.trim()}"${exportParams.query ? ` for "${exportParams.query}"` : ""} is being generated.`,
+          actions: [
```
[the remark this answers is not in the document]
rubinaga: fixed

src/features/activities/api/useExportActivitiesCsv.ts
```diff
@@ -0,0 +1,37 @@
...
+interface CreateActivitiesExportJobParams {
+  readonly name: string;
+  readonly query: string;
+  readonly selected_activity_ids: number[];
+  readonly selected_field_ids: string[];
+  readonly retain_until: string;
+}
+
+export const useExportActivitiesCsv = () => {
+  const authFetch = useFetch();
```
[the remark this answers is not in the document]
rubinaga: fixed

src/features/activities/components/ActivitiesExportForm/ActivitiesExportForm.tsx
```diff
@@ -0,0 +1,296 @@
...
+} from "./types";
+
+interface ActivitiesExportFormProps {
+  readonly exportParams: ActivityListParams;
+  readonly activityCount: number | undefined;
+  readonly selectedActivityCount?: number;
+  readonly selectedActivityIds?: number[];
+}
+
+const ActivitiesExportForm: FC<ActivitiesExportFormProps> = ({
```
yurii-vasyliev, line 24: **Form duplication.** This component and `InstancesExportForm.tsx` are ~90% identical — `toggleField`, `toggleGroupSelect`, `filteredFieldGroups`, `accordionSections`, and the entire step-0/step-1 render block are copy-pasted, and the two `.module.scss` files are byte-for-byte identical (same blob hash). Consider extracting a shared `ExportForm` in `@/features/exports` parameterized by `fieldGroups`, `initialValues`, the mutation hook, and the notification strings, leaving each feature a thin wrapper. Worth doing now, before the two ~300-line copies drift.

src/features/activities/components/ActivitiesExportForm/ActivitiesExportForm.tsx
```diff
@@ -0,0 +1,296 @@
...
+  exportParams,
+  activityCount: _activityCount,
+  selectedActivityCount: _selectedActivityCount,
```
yurii-vasyliev: **Dead props.** `activityCount` and `selectedActivityCount` are received and immediately aliased to unused `_`-prefixed names — `ActivitiesActions` passes them but they're never read here. Same pattern in `InstancesExportForm.tsx` (`instanceCount` / `selectedInstanceCount`). Either drop them from the props and call sites, or use them.

src/features/activities/api/useExportActivitiesTsv.ts
```diff
@@ -0,0 +1,37 @@
+import useFetch from "@/hooks/useFetch";
+import type { ApiError } from "@/types/api/ApiError";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import type { AxiosError, AxiosResponse } from "axios";
+import type { ActivitiesExportJob } from "../types/ActivitiesExportJob";
+
+interface CreateActivitiesExportJobParams {
+  readonly name: string;
+  readonly query: string;
+  readonly selected_activity_ids: number[];
```
yurii-vasyliev: **Inconsistent export contract.** This endpoint receives both `selected_activity_ids` and a `query` that already encodes the same selection (`id:1 OR id:2`, from `buildActivityExportQuery`), so the selection is sent twice. Meanwhile the instances endpoint (`useExportInstancesTsv`) sends no `selected_instance_ids` at all — query only. Please confirm the canonical contract with the backend and make the two consistent; if the query already encodes the selection, this field looks redundant.

src/features/instances/components/InstanceList/InstanceList.tsx
```diff
@@ -287,7 +299,7 @@ const InstanceList = memo(function InstanceList({
         },
       },
       {
-        accessor: "last_ping",
+        accessor: "last_ping_time",
```
yurii-vasyliev, line 356: **Possible breaking change for persisted column prefs.** Renaming the accessor `last_ping` → `last_ping_time` aligns it with the export field id, but `disabledColumns` is persisted as a URL param (`libs/pageParamsManager`). Existing bookmarked/shared URLs containing `disabledColumns=last_ping` will silently stop hiding this column after the rename. Probably acceptable, but it's an unrelated user-visible side effect riding along in an exports PR — worth a note in the changeset or a fallback mapping.

src/features/exports/components/ExportsListActions/ExportsListActions.tsx
```diff
@@ -0,0 +1,150 @@
...
+    setFalse: closeCancelModal,
+  } = useBoolean();
+
+  const handleDownload = async () => {
+    try {
+      const result = await onDownload(job);
+      if (result) {
+        notify.success({
+          title: "TSV download started",
+          message: `${job.name} has been downloaded and removed from the export list.`,
```
yurii-vasyliev, line 43: **Message asserts a deletion the client doesn't perform.** This tells the user the export was "removed from the export list", but the handler only invalidates queries — there's no delete call. It relies on the backend deleting the job on first download (the same wording is in `ExportDetailsSidePanel.tsx`). Please confirm that's the intended contract; stating "download = permanent removal" as fact is surprising UX if it isn't actually happening.

src/features/exports/components/ExportsContainer/ExportsContainer.tsx
```diff
@@ -0,0 +1,65 @@
...
+import { Link } from "react-router";
+import { useGetAllExportJobsList } from "../../api/useGetAllExportJobsList";
+import ExportsList from "../ExportsList";
+import { EXPORT_TYPE_OPTIONS } from "./constants";
+
+const ExportsContainer: FC = () => {
+  const { exportJobs, totalCount, isLoading } = useGetAllExportJobsList();
+
+  const { totalCount: unfilteredCount, isLoading: isLoadingUnfiltered } =
+    useGetAllExportJobsList({}, { listenToUrlParams: false });
```
yurii-vasyliev: **Double polling on this page (minor).** `useGetAllExportJobsList` is mounted twice — the filtered list plus this unfiltered count for empty-state detection — and `useGetExportJob` polls too, all with `refetchIntervalInBackground: true`. That's 2–3 background pollers firing every 5s even when the tab is unfocused. Consider whether the unfiltered count query needs to poll at all.

src/features/exports/types/ExportJob.ts
```diff
@@ -0,0 +1,18 @@
...
+  id: string;
+  name: string;
+  filename: string;
+  type: ExportJobType;
+  row_count: number;
+  created_at: string;
+  status: ExportJobStatus;
+  progress: number;
+  estimated_seconds_remaining?: number | null;
+  download_ready: boolean;
```
yurii-vasyliev, line 15: **Dead field (minor).** `download_ready` is set in every mock and MSW handler but never read in component logic — all UI branches on `status === "completed"`. Drop it, or wire it up if it's meant to gate the download button.

src/features/activities/components/ActivitiesExportForm/constants.ts
```diff
@@ -0,0 +1,88 @@
...
+];
+
+export const INITIAL_VALUES: ActivitiesExportFormValues = {
+  name: "",
+  selectedFieldIds: VISIBLE_COLUMN_FIELD_IDS,
+  retainUntil: moment().add(3, "years").format(INPUT_DATE_FORMAT),
+};
+
+export const VALIDATION_SCHEMA = Yup.object().shape({
+  name: Yup.string().trim().required("This field is required."),
```
yurii-vasyliev: **Validation message inconsistency (minor).** These use trailing periods ("This field is required.", "Select at least one attribute.") while `InstancesExportForm/constants.ts` omits them, on otherwise-identical side-by-side forms. Pick one for consistency.

src/features/activities/components/ActivitiesExportForm/helpers.ts
```diff
@@ -0,0 +1,12 @@
+export const buildActivityExportQuery = ({
```
yurii-vasyliev: **Duplicate id-query helper (minor).** `buildActivityExportQuery` and instances' `buildExportQuery` produce the same `id:… OR id:…` string two different ways (`.map(id => \`id:${id}\`).join(" OR ")` vs `id:${ids.join(" OR id:")}`). Fold into one shared helper in `@/features/exports`.

src/features/instances/api/useExportInstancesTsv.ts
```diff
@@ -0,0 +1,36 @@
+import useFetch from "@/hooks/useFetch";
+import type { ApiError } from "@/types/api/ApiError";
+import { useMutation, useQueryClient } from "@tanstack/react-query";
+import type { AxiosError, AxiosResponse } from "axios";
+import type { InstancesExportJob } from "../types/InstancesExportJob";
+import type { InstanceListParams } from "../helpers";
+
+interface CreateInstancesExportJobParams extends InstanceListParams {
```
yurii-vasyliev, line 8: **Minor consistency.** `CreateInstancesExportJobParams` fields aren't `readonly`, while `CreateActivitiesExportJobParams` marks them `readonly`. Align the two.

yurii-vasyliev (commented):
Inline notes from the PR review — significant findings plus a few minor cleanups. Nothing blocking; the backend-contract questions (dead props, redundant `selected_activity_ids`, and the "removed from the export list" message) are the ones I'd resolve before merge.

src/features/exports/components/SortableFieldList/SortableFieldList.tsx
marqode, line 1: When using this component (macOS, google chrome) I see a globe icon slide to the mouse position whenever I start to drag a column in the list.

https://github.com/user-attachments/assets/746c85ff-0ea8-407a-90db-ee0c29da19d7

src/features/exports/components/ExportDetailsSidePanel/ExportDetailsSidePanel.tsx
```diff
@@ -0,0 +1,208 @@
...
+        title: "TSV discarded",
+        message: `${job.name} has been discarded.`,
+      });
+    } catch (error) {
+      debug(error);
+    }
+  };
+
+  if (!job) {
+    return <SidePanel.LoadingState />;
```
marqode, line 20: When the job is not found the panel shows loadingState forever; should display something more informative if the API returns 400/500

Also could be worth adding a mock job with id `job-new` to test creating an export -> linking to Exports page with MSWs

src/tests/server/handlers/exports.ts
```diff
@@ -0,0 +1,101 @@
...
+      if (status === "empty") {
+        return HttpResponse.json({ count: 0, results: [] });
+      }
+      if (status === "variant" && Array.isArray(response)) {
+        return HttpResponse.json(
+          generateExportJobsResponse(request, response as ExportJob[]),
+        );
+      }
+    }
+    return HttpResponse.json({ count: 0, results: [] });
```
marqode: Non-blocking but it'd be nice to return the mocked jobs here for testing the UI with MSWs

src/tests/server/handlers/exports.ts
```diff
@@ -0,0 +1,101 @@
...
+  }),
+
+  http.get(`${API_URL}${EXPORTS_PATH}/:jobId`, () => {
```
marqode:
```suggestion
  http.get(`${API_URL}${EXPORTS_PATH}/:jobId`, ({ params }) => {
```
Same as above, loading job details in side panel would help with testing the UI

src/tests/server/handlers/exports.ts
```diff
@@ -0,0 +1,101 @@
...
+      if (
+        status === "variant" &&
+        response &&
+        typeof response === "object" &&
+        !Array.isArray(response)
+      ) {
+        return HttpResponse.json(response as unknown as ExportJob);
+      }
+    }
+    return new HttpResponse(null, { status: 404 });
```
marqode:
```suggestion
    const job = DEFAULT_EXPORT_JOBS.find((j) => j.id === params.jobId);
    if (!job) return new HttpResponse(null, { status: 404 });
    return HttpResponse.json(job);
```

src/features/exports/components/ExportsListActions/ExportsListActions.tsx
```diff
@@ -0,0 +1,150 @@
...
+        message: `${job.name} has been discarded.`,
+      });
+    } catch (error) {
+      debug(error);
+    } finally {
+      closeDiscardModal();
+    }
+  };
+
+  const downloadActions =
```
marqode: Should we support a Retry action for failed downloads? It feels intuitive that there would be an option to attempt to export again with the same options, or show an error message / reason the export failed that would block a retry.

Image

src/features/exports/components/ExportsList/ExportsList.tsx
```diff
@@ -0,0 +1,125 @@
...
+      },
+      {
+        Header: "Created",
+        accessor: "job",
+        id: "createdAt",
+        Cell: ({ row }: CellProps<ExportRowData>) => (
+          <>
+            {moment(row.original.job.created_at).format(
+              DISPLAY_DATE_TIME_FORMAT,
+            )}
```
marqode, line 86: Super minor but the dateTime columns in /instances use a different font

Image

src/features/exports/api/useGetExportJob.ts
```diff
@@ -0,0 +1,30 @@
+import useFetch from "@/hooks/useFetch";
+import type { ApiError } from "@/types/api/ApiError";
+import { useQuery } from "@tanstack/react-query";
+import type { AxiosError, AxiosResponse } from "axios";
+import { EXPORT_JOBS_POLL_INTERVAL_MS } from "../constants";
+import type { ExportJob } from "../types/ExportJob";
+
+export const useGetExportJob = (id: string) => {
+  const authFetch = useFetch();
+
```
marqode, line 13: add a guard like `enabled: !!id,` to the query options in case id is null, eg when navigating to `/exports?sidePath=view`

marqode (changes requested):
Looking good! The non-null id guard in `useGetExportJob` feels the most pressing, the changes to MSWs are just so that exports can be tested more easily in the UI.

Outside the scope of this PR, but I haven't been able to download any exports when running this branch against feature/tsv-exports in landscape-server. API endpoints like `/api/v2/exports/119/download` give me 404s like this one:
```
{
    "error": "ApiRequestError",
    "message": "Export `119` is no longer available.",
    "detail": null
}
```

src/features/exports/components/ExportsListActions/ExportsListActions.tsx
rubinaga: added a retry option here, nice suggestion

src/features/instances/components/InstanceList/InstanceList.tsx
rubinaga: the accessor should be `last_ping_time` as that's the field in the object, so I'm keeping this as is

src/features/exports/components/SortableFieldList/SortableFieldList.tsx
rubinaga: I can't reproduce this in my own google chrome in ubuntu, but I tried adding a fix. Let me know if it happens again

marqode: that fixed it!

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+        className={classes.bar}
+        role="progressbar"
+        aria-valuenow={clampedProgress}
+        aria-valuemin={0}
+        aria-valuemax={MAX_PROGRESS}
+        aria-label="Progress"
+      >
+        <div
+          className={classes.fill}
+          style={{ width: `${clampedProgress}%` }}
```
marqode: This is a nit but our `copilot-instructions:110` has "- Reject inline styling or direct DOM manipulation." so it's possible this would get flagged at some point.

marqode (dismissed):
LGTM! One comment about inline style for progress bar but it may be fine to leave as-is if we use that same pattern elsewhere

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,65 @@
...
+import classNames from "classnames";
+
+export interface ProgressBarProps {
+  readonly progress: number;
+  readonly secondsRemaining: number | null;
+  readonly fullWidth?: boolean;
+  readonly loading?: boolean;
+}
+
+const ProgressBar: FC<ProgressBarProps> = ({
```
marqode, line 15: @gesquivelgaghi also created a ProgressBar component [in this PR](https://github.com/canonical/landscape-ui/pull/691/changes#diff-b0e8fb2bb332a087b37d57804d669f32e39beeffe19f89a9814ab1fc4b72f7de). Once both features are merged they should use the same component.

src/features/instances/components/InstancesExportForm/InstancesExportForm.tsx
```diff
@@ -0,0 +1,99 @@
...
+interface InstancesExportFormProps {
+  readonly exportParams: InstanceListParams;
+  readonly selectedInstanceIds?: number[];
+}
+
+const InstancesExportForm: FC<InstancesExportFormProps> = ({
+  exportParams,
+  selectedInstanceIds,
+}) => {
+  const { closeSidePanel } = useSidePanel();
```
marqode, line 30: Is there a reason `ExportForm` uses the side panel directly instead of adding to `pageParams`? Not a blocker here, but for the reports export closing the form should return them to the original report view. If that feels out of scope for this PR I'm OK with adding it in mine.

src/features/exports/components/ExportDetailsSidePanel/ExportDetails.tsx
```diff
@@ -0,0 +1,225 @@
...
+        });
+      }
+    } catch (error) {
+      debug(error);
+    }
+  };
+
+  const handleConfirmCancel = async () => {
+    try {
+      await onCancel(job.id);
```
marqode, line 55:
```suggestion
    await onCancel(job.id);
    popSidePathUntilClear();
```

src/features/exports/components/ExportDetailsSidePanel/ExportDetails.tsx
```diff
@@ -0,0 +1,225 @@
...
+      });
+    } catch (error) {
+      debug(error);
+    }
+  };
+
+  const handleRetry = async () => {
+    try {
+      const { data: newJob } = await onRetry(job.id);
+      setPageParams({ name: String(newJob.id) });
```
marqode, line 82: Do we also need to remove the old job from the list here? Or is the intention that it continues to show Failed for the original job until the user removes it?

src/features/activities/components/ActivitiesActions/ActivitiesActions.tsx
```diff
@@ -66,6 +116,14 @@ const ActivitiesActions: FC<ActivitiesActionsProps> = ({ selected }) => {
   return (
     <div key="buttons" className="p-segmented-control">
       <div className="p-segmented-control__list">
+        <Button
+          className="p-segmented-control__button"
+          type="button"
+          disabled={!isAllSelected && selected.length === 0}
+          onClick={handleExport}
+        >
+          <span>Export selection as TSV</span>
```
marqode, line 129: Right border of button remains grayed out when enabled. Might be a better fit for an upstream fix.

Image

src/components/ProgressBar/ProgressBar.tsx
```diff
@@ -0,0 +1,43 @@
...
+
+  return (
+    <div className={`${classes.wrapper} ${fullWidth ? classes.fullWidth : ""}`}>
+      <div
+        className={classes.bar}
+        role="progressbar"
+        aria-valuenow={clampedProgress}
+        aria-valuemin={0}
+        aria-valuemax={MAX_PROGRESS}
+        aria-label="Progress"
```
marqode:
```suggestion
        aria-label={label || "Progress"}
```
"Progress" label is a bit generic, consider adding optional label prop to help differentiate from other progress bars on the same page.

src/features/activities/components/ActivitiesActions/helpers.ts
```diff
@@ -0,0 +1,19 @@
+import { pluralize } from "@/utils/_helpers";
+
+export const getExportTitle = ({
```
marqode: Consider refactoring this helper to feature/exports so that it can be used for other export types.

marqode (dismissed):
A few suggestions / questions but nothing blocking

src/features/activities/components/ActivitiesActions/helpers.ts
rubinaga: refactored it

src/components/ProgressBar/ProgressBar.tsx
rubinaga: Good point, added an optional label

src/features/activities/components/ActivitiesActions/ActivitiesActions.tsx
rubinaga: Yeah I think it should be upstream

src/features/exports/components/ExportDetailsSidePanel/ExportDetails.tsx
rubinaga: On retry the old job is removed in the backend and the new one takes its place, we invalidate the table cache to fetch the new data

src/features/exports/components/ExportDetailsSidePanel/ExportDetails.tsx
rubinaga: good point, added

src/features/instances/components/InstancesExportForm/InstancesExportForm.tsx
rubinaga: In this case it's because the whole `InstancesPageActions.tsx` component uses `useSidePanel`. I just kept it like this for now since I saw no difference

src/components/ProgressBar/ProgressBar.tsx
rubinaga: Yeah I will keep an eye out for it

yurii-vasyliev (dismissed):
LGTM, good job!

marqode (approved):
LGTM!

yurii-vasyliev (approved):
LGTM!

PR #688, opened 2026-06-29
refactor: replace moment.js with custom class

rubinaga:
## Summary

Replaces the `moment` dependency with a custom `LandscapeDate` class (`src/libs/date/LandscapeDate.ts`) to reduce bundle size. The new class preserves moment's parsing, formatting, UTC/local, calendar, and diff behavior, including DST adjustment.

Test coverage: 43 unit tests in `LandscapeDate.test.ts` (including a DST cross-over case). Full suite: 3147/3147 passing.

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/src/components/form/DeliveryScheduling/components/DeliveryBlock/DeliveryBlock.tsx
+++ b/src/components/form/DeliveryScheduling/components/DeliveryBlock/DeliveryBlock.tsx
@@ -1,5 +1,5 @@
 import type { FormikContextType } from "formik";
-import moment from "moment";
+import date from "@/libs/date";
 import { Input } from "@canonical/react-components";
 import { getFormikError } from "@/utils/formikErrors";
 import RadioGroup from "@/components/form/RadioGroup";
@@ -34,7 +34,7 @@ const DeliveryBlock = <T extends DeliveryProps>({
           onSelect: async () => {
             await formik.setFieldValue(
               "deliver_after",
-              moment().add(5, "minutes").toISOString().slice(0, 16),
+              date().add(5, "minutes").toISOString().slice(0, 16),
             );
           },
           expansion: (
--- a/src/libs/date/ChronoDate.ts
+++ b/src/libs/date/ChronoDate.ts
@@ -1,10 +1,12 @@
 /**
- * LandscapeDate is a small wrapper around the native `Date` object. It
- * centralizes the date behavior and call style the Landscape UI relies on.
+ * ChronoDate is a small, dependency-free wrapper around the native `Date`
+ * object. It centralizes date behavior behind a small, consistent call style.
  *
- * It intentionally implements only the subset used in this codebase, delegating
- * to native `Date` for arithmetic, comparison and serialization and
+ * It intentionally implements only a focused subset of date operations,
+ * delegating to native `Date` for arithmetic, comparison and serialization and
  * adding a small token-based formatter that the native APIs do not provide.
+ *
+ * Locale is English-only by design.
  */
 
 const MONTHS_SHORT = [
@@ -126,7 +128,7 @@ export interface CalendarFormats {
 }
 
 export type DateInput =
-  | LandscapeDate
+  | ChronoDate
   | Date
   | string
   | number
@@ -310,7 +312,7 @@ const FORMAT_TOKENS: Record<string, (fields: DateFields) => string> = {
   a: (f) => (f.H < HOURS_PER_HALF_DAY ? "am" : "pm"),
 };
 
-export class LandscapeDate {
+export class ChronoDate {
   private readonly _time: number;
   private readonly _utc: boolean;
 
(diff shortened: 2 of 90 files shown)


marqode (commented):
Agentic review summary

### Verified clean
- Zero remaining `moment` references in `src/`; dep removed from `package.json`; lockfile consistent.
- LSP clean on `src/libs/date/`; no `console.log`/`as any`/`@ts-ignore`/TODO added.
- Format-token parity: all format strings in use (`YYYY-MM-DD`, `YYYY-MM-DDTHH:mm:ss`, `MMM D, YYYY`, `MMM DD, YYYY, HH:mm`, `[Z]` literal) are implemented by `FORMAT_TOKEN_REGEX`.
- Immutability respected: every `.add()`/`.subtract()` call site uses the returned instance (moment-mutation footgun absent).
- DST day-diff math in `diff(_, "days")` verified correct; `calendar()` bucket ranges match moment; month/year rollover arithmetic verified.
- Tests: 47/47 `ChronoDate` tests, plus 81/81 tests across all changed feature test files.

### Warnings (fix or acknowledge)
1. **`src/libs/date/ChronoDate.ts:566`** — `toISOString()` returns `""` on invalid dates; moment returned `null`. Currently unreachable: the three `date(values.retainUntil).toISOString()` API-payload call sites (`ActivitiesExportForm.tsx:52`, `InstancesExportForm.tsx:56`, `ReportExportForm.tsx:105`) sit behind Yup `.required()` + future/max-date tests, so Formik never submits an invalid value. Library's public contract now diverges from moment's; any future caller piping an optional date straight to a payload will send `""` where the backend previously received `null`. Suggest returning `null` (and typing `string | null`) and adjusting the two `.toISOString().slice(0, 16)` inits (moment actually *threw* there via `null.slice` — ChronoDate is safer as-is). 
*Alternative*: keep `""` and document the divergence in the module doc.
2. **`src/libs/date/ChronoDate.ts:279-298`** — Dead code: `ISO_8601_PARTS_REGEX` (all groups optional) already matches date-only and no-timezone inputs, so the `DATE_ONLY_REGEX` and `DATE_TIME_NO_TZ_REGEX` fallback branches in `parseString` are unreachable. Harmless, but misleading for future maintenance — remove or reorder.
3. **`src/libs/date/ChronoDate.ts:181`** — `hasValidDateTimeParts` caps hour at 23; ISO 8601 permits `24:00:00` (end-of-day midnight) which moment accepts. Edge case; likely never seen from the API, but a strict-parity note.

### Nitpicks (non-blocking)
- `ChronoDate.ts:146` `to12Hour` condenses to `(h % 12) || 12`.
- `ChronoDate.ts:211` offset validation accepts up to `+23:59`; ISO caps at `+14:00` (harmless delegation).
- Export-form payload mappings rely implicitly on Yup gating rather than an explicit `.isValid()` check.

The migration is semantically clean at every reachable call site, test coverage is solid, and the riskiest diff patterns (immutability, format tokens, DST diff) all check out.

src/libs/date/ChronoDate.ts
```diff
@@ -0,0 +1,637 @@
...
+  isSameOrAfter(other: DateInput): boolean {
+    return this._time >= ChronoDate.from(other)._time;
+  }
+
+  isSameOrBefore(other: DateInput): boolean {
+    return this._time <= ChronoDate.from(other)._time;
+  }
+
+  toISOString(): string {
+    return this.isValid() ? new Date(this._time).toISOString() : "";
```
marqode: consider returning `null` here to match moment.js behavior

src/libs/date/ChronoDate.ts
```diff
@@ -0,0 +1,637 @@
...
+    return createLocalTime(
+      Number(y),
+      Number(m),
+      Number(d),
+      Number(h),
+      Number(min),
+      Number(s ?? 0),
+      parseMilliseconds(ms),
+    );
+  }
```
marqode: Unreachable code in parseString. The ISO_8601_PARTS_REGEX marks all time and timezone parts as optional (?:...)?. Because of this, it will successfully match date-only strings like "2024-03-15" and no-timezone strings like "2024-03-15T10:30:00". As a result, the subsequent fallback checks for DATE_ONLY_REGEX and DATE_TIME_NO_TZ_REGEX are dead code and will never be executed.

src/libs/date/ChronoDate.ts
```diff
@@ -0,0 +1,637 @@
...
+  hour = 0,
+  minute = 0,
+  second = 0,
+): boolean =>
+  month >= 1 &&
+  month <= MONTHS_PER_YEAR &&
+  day >= 1 &&
+  day <= getDaysInMonth(year, month) &&
+  hour >= 0 &&
+  hour <= MAX_HOUR &&
```
marqode: valid-ISO 24:00:00 rejected (moment accepts)

marqode (approved):
LGTM, a few warnings flagged that could be worth addressing before its merged

ethanashaw (approved):
I saw a few lines of missing test coverage in `ChronoDate.ts` in case they are feasible to cover, overall looks good

src/components/form/DeliveryScheduling/components/DeliveryBlock/DeliveryBlock.tsx
```diff
@@ -34,7 +34,7 @@ const DeliveryBlock = <T extends DeliveryProps>({
           onSelect: async () => {
             await formik.setFieldValue(
               "deliver_after",
-              moment().add(5, "minutes").toISOString().slice(0, 16),
+              date().add(5, "minutes").toISOString()?.slice(0, 16) ?? "",
             );
```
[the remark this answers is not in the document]
marqode: agree with this comment and the similar one in `HoldSnapForm.tsx`. Consider replacing with `date().add(5, "minutes").format("YYYY-MM-DDTHH:mm")`

`ChronoDate.format()` implementation explicitly uses local date fields unless `.utc()` has been called, so this preserves the intended wall-clock value.

marqode (commented):
agree with the copilot suggestions for DeliveryBlock and HoldSnapForm. Each of those components could get a new test case to confirm date uses the correct local-formatted value.

ethanashaw (approved):

yurii-vasyliev (approved):

PR #694, opened 2026-06-30
refactor: move users to features

rubinaga:
## Summary

(Low priority)

Refactored users to be inside the feature component and use the api component, as well as small things like moving out things to constants or helper files.

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


src/pages/dashboard/instances/[single]/tabs/users/UserPanel/UserPanel.tsx
```diff
@@ -1,123 +1,8 @@
...
-        users={users}
-      />
-      <TablePagination
-        handleClearSelection={handleClearSelection}
-        totalItems={filteredUsers.length}
-        currentItemCount={users.length}
-      />
-    </>
-  );
+  return <UserContainer />;
```
Comment (gesquivelgaghi) [inline src/pages/dashboard/instances/[single]/tabs/users/UserPanel/UserPanel.tsx:5]: Do we really need this component that is literally only a wrapper for another component? Cant we cut out the middleman here and just call `UserContainer` directly where this is called, or delete `UserContainer` and put its contents here?

src/features/users/components/UsersEmptyState/UsersEmptyState.tsx
```diff
@@ -0,0 +1,34 @@
+import EmptyState from "@/components/layout/EmptyState";
+import { Button } from "@canonical/react-components";
+import type { FC } from "react";
+import {
+  ADD_USER_LABEL,
+  EMPTY_STATE_BODY,
+  EMPTY_STATE_TITLE,
+} from "./constants";
```
gesquivelgaghi: Why do these need to be constants? Cant we just use the strings directly in this file?

src/features/users/api/useGetUsers.ts
```diff
@@ -0,0 +1,39 @@
...
+import type { ApiPaginatedResponse } from "@/types/api/ApiPaginatedResponse";
+import type { User } from "@/types/User";
+import { useQuery } from "@tanstack/react-query";
+import type { AxiosError, AxiosResponse } from "axios";
+
+export interface GetUsersParams {
+  computer_id: number;
+  limit?: number;
+  offset?: number;
+  query?: string;
```
marqode, line 12: Old `useUsers.ts` stripped `query: queryParams?.query || undefined` before hitting query key/fetch. This hook drops that. No caller sends `query` today but keeping the field here means a future caller could send the `?query=""` empty param to API. Remove field til wired, or restore strip logic.

src/features/users/api/useGetGroups.ts
```diff
@@ -0,0 +1,33 @@
...
+import type { ApiError } from "@/types/api/ApiError";
+import type { GroupsResponse } from "@/types/User";
+import { useQuery } from "@tanstack/react-query";
+import type { AxiosError, AxiosResponse } from "axios";
+
+export interface GetGroupsParams {
+  computer_id: number;
+}
+
+export const useGetGroups = (params: GetGroupsParams) => {
```
marqode, line 11: `useUsers()` query hooks previously took `config` options (e.g. `enabled`), matching the patternin `copilot-instructions.md`. New hooks (`useGetUsers`, `useGetGroups`, `useGetUserGroups`) take none. 

Not currently needed by any callers so feel free to ignore if this was intentional.

marqode (dismissed):
LGTM!
Two non-blocking notes below (dead `query` field no longer stripped, dropped `config` passthrough) for optional follow-up.

PR #695, opened 2026-06-30
Chore: Remove duplicated api calls in mirrors

gesquivelgaghi:
## Summary

Summarize the changes made in this pull request. Include any relevant context or background information that would help reviewers understand the purpose and scope of the changes.

## Release Impact

Change type (tick one):

- [ ] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


src/features/publications/components/AssociatedPublicationsCount/AssociatedPublicationsCount.test.tsx
```diff
@@ -0,0 +1,36 @@
...
+  });
+
+  it("renders publication count as link when loaded", async () => {
+    renderWithProviders(
+      <AssociatedPublicationsCount sourceName={repositories[0].name} />,
+    );
+
+    expect(await screen.findByRole("link")).toHaveAttribute(
+      "href",
+      expect.stringContaining(encodeURIComponent(repositories[0].name)),
```
marqode:
```suggestion
      expect.stringContaining(`source%3A${encodeURIComponent(repositories[0].name)}`)
```
slightly tighter assertion to include `source` in string

src/features/mirrors/components/MirrorsList/MirrorsList.tsx
```diff
@@ -115,9 +115,7 @@ const MirrorsList: FC<MirrorsListProps> = ({ mirrors, emptyMsg }) => {
         Header: "Publications",
         Cell: ({ row: { original: mirror } }: CellProps<Mirror>) =>
           mirror.name ? (
-            <Suspense fallback={<LoadingState inline />}>
-              <MirrorPublicationsLink mirrorName={mirror.name} />
-            </Suspense>
+            <AssociatedPublicationsCount sourceName={mirror.name} />
```
marqode, line 118: This might be intentional so take it or leave it:
The old `useSuspenseQuery` would throw to an error boundary; `useGetPublicationsBySource` returns `[]` on error, so a broken API call renders "0 publications" with no user-visible error message.

marqode (dismissed):
LGTM, one suggestion and a comment on what may be a slight regression in error reporting.

src/features/mirrors/components/PublishMirrorForm/PublishMirrorForm.tsx
```diff
@@ -5,50 +5,47 @@ import SidePanel from "@/components/layout/SidePanel";
...
-            disabled={!publicationTargets.length}
-          />
-          <RadioInput
-            label="Existing publication"
-            checked={!useNewPublication}
-            onChange={toggle}
-            disabled={!publications.length}
-          />
-        </div>
+        {!!publications.length && (
```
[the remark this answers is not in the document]
gesquivelgaghi: Neither option works when there are no publication targets, which is why the buttons to open this side panel redirect to a "no publication targets" modal instead when thats the case, and do not open this side panel.

src/features/mirrors/components/MirrorsList/MirrorsList.tsx
gesquivelgaghi: what if i add this to `useGetPublicationsBySource` before it returns? where `error` comes from the API call
```
if (error) {
    throw new Error(error.message);
  }
```

marqode: If we want to handle this error I think it'd be cleanest to include it in the `useGetPublicationsBySource`return type so it doesn't crash other components
```
return {
  publications: data ?? [],
  isGettingPublications: isLoading,
  isPublicationsError: isError,
};
```
then use it to display `NO_DATA_TEXT` or similar in `AssociatedPublicationsCount`:
```
const { publications, isGettingPublications, isPublicationsError } =
  useGetPublicationsBySource(sourceName);

if (isGettingPublications) return <LoadingState inline />;
if (isPublicationsError) return <NoData />;
```

gesquivelgaghi: If we dont want to have the potential API error show an error screen then i'd just leave it as is to show `0 instances` tbh, I feel like the difference between that and `---` isnt significant enough to the user experience.

marqode (approved):
LGTM

rubinaga (approved):
LGTM

PR #699, opened 2026-07-02
chore: remove unused getStrippedUrl helper

yurii-vasyliev:
## What

Removes the unused `getStrippedUrl` helper from `AddMirrorForm/helpers.ts`.

## Why

`getStrippedUrl` stripped the Ubuntu Pro bearer token from a mirror's source URL on the client. The server now strips the token itself, so the URL already arrives stripped — the client-side helper is unnecessary. It currently has **no callers** anywhere in the codebase (dead code).

This supersedes #667, which would have wired `getStrippedUrl` into the mirror source URL display; that approach is obsolete now that stripping happens server-side.

Empty changeset — internal cleanup, no user-facing change.


marqode (approved):
LGTM

gesquivelgaghi (approved):

PR #700, opened 2026-07-02
Fix/lndeng 4566 instances empty state

RyanLoi98:
## Summary

Added empty state for when there are no instances registered to Landscape.

## Release Impact

Change type (tick one):

- [X] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>


src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,44 @@
...
+
+  return (
+    <EmptyState
+      title="No instances found"
+      icon="connected"
+      body="You don't have any instances registered to Landscape yet."
+      link={{
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/manage-computers/",
+        text: "How to manage instances in Landscape",
+      }}
```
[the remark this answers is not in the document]
RyanLoi98: this link is the new updated link. I will fix the test to address the change in the link.

src/pages/dashboard/instances/InstancesContainer/helpers.ts
```diff
@@ -0,0 +1,21 @@
...
+  instanceCount: number | undefined,
+  isFilteringInstances: boolean,
+  isInstanceLoading: boolean,
+) =>
+  currentPage === 1 &&
+  pageSize === DEFAULT_PAGE_SIZE &&
+  instanceCount !== undefined &&
+  !isInstanceLoading &&
+  instanceCount === 0 &&
+  !isFilteringInstances;
```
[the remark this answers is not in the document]
RyanLoi98: We don't have a situation where there are no instances and the instances page will load on anything but page 1, or have a non default page size. This is a defined behavior in our code base.

src/features/instances/components/InstanceList/InstanceList.tsx
```diff
@@ -361,11 +361,7 @@ const InstanceList = memo(function InstanceList({
   return (
     <ResponsiveTable
       subhead={subhead}
-      emptyMsg={
-        isFilteringInstances
-          ? "No instances found according to your search parameters."
-          : "No instances found"
-      }
+      emptyMsg="No instances found according to your search parameters."
```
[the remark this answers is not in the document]
RyanLoi98: Ah yes I see, I removed the unused function and filters variable coming from usePageparams. This will fix the issue.

src/features/instances/api/useGetInstances.ts
```diff
@@ -73,6 +74,7 @@ export const useGetInstances = (
     isErrorInstances: isError,
     isFetchingInstances: isFetching,
     isGettingInstances: isPending,
+    isInstanceLoading: isLoading,
```
gesquivelgaghi: why not just use `isGettingInstances`?

RyanLoi98: I thought this would be a more robust way of checking to see if instances were loading, and was in line with how we do it on the scripts side of things. 

this is because isloading = isFetching && isPending. So it encapsulates the behavior of isGettingInstances while also ensuring we check if the fetch query is still in process of executing.

gesquivelgaghi: Interesting that its what we use for scripts. In general in the codebase we tend to default to `isPending` since in the vast majority of cases we dont want to override cached data with a loading state before updating the data. The smoother user experience is to keep the old value until the new one is fully fetched. I think that applies here too.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,44 @@
...
+                key="register-instance"
+                onClick={() => {
+                  window.location.assign(registerUrl);
+                }}
+                type="button"
+              >
+                Register instance
+              </Button>,
+            ]
+          : []
```
gesquivelgaghi: nit-pick, but since its an optional field it makes more sense to leave it undefined when there's no value  
```suggestion
          : undefined
```

src/pages/dashboard/instances/InstancesContainer/helpers.ts
```diff
@@ -0,0 +1,21 @@
+import { DEFAULT_PAGE_SIZE } from "@/libs/pageParamsManager";
+
+export const isInstanceLoadingState = (
+  currentPage: number,
+  pageSize: number,
+  isInstanceLoading: boolean,
+) => currentPage === 1 && pageSize === DEFAULT_PAGE_SIZE && isInstanceLoading;
```
gesquivelgaghi: why do we need `currentPage === 1 && pageSize === DEFAULT_PAGE_SIZE` here?

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,44 @@
...
+      icon="connected"
+      body="You don't have any instances registered to Landscape yet."
+      link={{
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/manage-computers/",
+        text: "How to manage instances in Landscape",
+      }}
+      cta={
+        registerUrl
+          ? [
+              <Button
```
gesquivelgaghi: oh this should be a link. you can style links as buttons using `className="p-button--positive"`

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,44 @@
+import type { FC } from "react";
+import { Button } from "@canonical/react-components";
+import EmptyState from "@/components/layout/EmptyState";
+import { useAuthHandle } from "@/features/auth";
+
+const REGISTER_PATH = "/how-to-register";
+
+const InstanceEmptyState: FC = () => {
+  const { getClassicDashboardUrlQuery } = useAuthHandle();
+  const { data } = getClassicDashboardUrlQuery();
```
gesquivelgaghi: i think we should get `isPending` here too and have the cta text content be the loading spinner until we have the url and then switch to the "Register instance" text. Or just anything to let the user know some information is being fetched

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,44 @@
...
+  const registerUrl = data?.data.url
+    ? `${data.data.url}${REGISTER_PATH}`
+    : undefined;
+
+  return (
+    <EmptyState
+      title="No instances found"
+      icon="connected"
+      body="You don't have any instances registered to Landscape yet."
```
gesquivelgaghi, line 9: I'd also run these copy and icon choices by Dani. I think nowadays we usually have `You don't have any [type] yet` as the title, and then some explanation about it as the body, but given how basic instances are, not sure what he'd want.

Also not sure the connected icon is the most appropriate imagery, since its not a connection issue or anything related to that. For the most recent pages we've made, we tend to go with no icon, but I think it's just good to check with Dani since its a design decision.

If you've already talked to him just disregard this

src/pages/dashboard/instances/InstancesContainer/helpers.ts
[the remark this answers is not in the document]
gesquivelgaghi: You're right, but what is the purpose of adding them here? It seems like it just complicates the logic. Also you dont need both of `instanceCount !== undefined` and `instanceCount === 0`, since the latter will always cover the first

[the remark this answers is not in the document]
RyanLoi98: yeah you're right, I just left that there because I noticed undefined could be returned from the backend for the count and just wanted to account for it. I will remove it.

src/pages/dashboard/instances/InstancesContainer/helpers.ts
RyanLoi98: That is an excellent question. I took this from the scripts helper, and I initially thought it was used because all instances were grabbed on page 1, then this prevents the loading state from appearing unnecessarily on subsequent pages. But this isn't the case, the api gets slices of data for each page but none of those subsequent pages will get a loading state as a result of this guard - which might not be good.

But I suspect this was done to prevent the loading spinner from coming on every page with every single page transition which looks like a quick annoying flash (or page size selection after first load), which makes sense. But I am actually not sure why this design principle was chosen. 

@yurii-vasyliev could you clarify why currentPage === 1 && pageSize === DEFAULT_PAGE_SIZE was needed (maybe I am missing something here)? I noticed you created the script version of this, which is why I am tagging you.

RyanLoi98: But it also brings up the question, should we have the spinner on every page switch? Because technically it is loading data from the back end so it should probably be indicating this for those with really slow internet.

src/features/instances/api/useGetInstances.ts
RyanLoi98: right I agree, but I don't think isLoading will cause premature data overwriting because the conditional still relies on the same isPending condition to be true. So whenever isPending is true isLoading will be as well, and the same for when it is false. So I still think isLoading is better, because it also checks to see if the query is still in process of executing and doesn't cause the empty state to trigger prematurely.

so if preserving cached data is the goal this might be the better approach.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
RyanLoi98: Makes sense to me!

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
RyanLoi98: yeah makes sense, it'll probably help with Aria too. I will be making that change on the scripts side too since it is a button over there.

RyanLoi98: Again great catches Gabe! It really helps me learn!

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
RyanLoi98: Sounds good, I'll send him a message. Again I just picked this up from the scripts side of thing, and it made sense to me since no instances is like not having any connected computers.

RyanLoi98:
Overall excellent PR review @gesquivelgaghi you really got me thinking about these things, and taught me alot more about the code base. It also helped me catch alot of bugs that I will subsequently fix on the scripts side of things. I really appreciate it!

src/features/instances/api/useGetInstances.ts
gesquivelgaghi: Oh you're right, I misinterpreted the `isLoading` condition as an `or` instead of `and` honestly because I couldn't really see the use of `and` lol. I looked into it a bit and I found this question being asked and answered here https://github.com/TanStack/query/discussions/6297 where a tanstack maintainer explains why using `isPending` is better in most cases.

src/pages/dashboard/instances/InstancesContainer/helpers.ts
gesquivelgaghi: I agree we should have it in every page whenever data is being fetched, and most features already only use `isPending` for the loading condition so it would be in line with what we do.

In general, when you're looking for code to use as a reference in this codebase I'd recommend you check when it was written, and if its over a year old you might wanna try to find a more recent implementation in a different feature and compare to see if we've changed how we do things. There's a lot of older parts of Landscape with slightly different standards, be that copy phrasing or code structure

src/features/instances/api/useGetInstances.ts
RyanLoi98: ok I have made the change back to isPending! (It'll be visible when I push the code)

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.tsx
```diff
@@ -26,14 +32,51 @@ const InstancesContainer = memo(function InstancesContainer({
...
+  const isFilteringInstances =
+    !!query ||
+    !!status ||
+    !!os ||
+    !!groupBy ||
+    !!contractExpiryDays ||
+    accessGroups.length > 0 ||
+    availabilityZones.length > 0 ||
+    tags.length > 0 ||
+    wsl.length > 0;
```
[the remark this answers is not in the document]
RyanLoi98: You're right. I will remove the groupBy variable from the filtering.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
RyanLoi98: Agreed. But I removed the button as per Danny's feedback so it's not needed anymore.

src/pages/dashboard/instances/InstancesContainer/helpers.ts
RyanLoi98: Sounds good, thanks for the advice and changes are now made.

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.tsx
```diff
@@ -26,6 +32,41 @@ const InstancesContainer = memo(function InstancesContainer({
...
+
+  if (
+    isInstancesEmptyState(
+      currentPage,
+      pageSize,
+      instanceCount,
+      isFilteringInstances,
+      isGettingInstances,
+    )
+  ) {
```
gesquivelgaghi:
```suggestion
  const hasNoInstances = !isGettingInstances && instanceCount === 0 && !isFilteringInstances;

  if (hasNoInstances) {
```

this doesnt really have to be a helper function, its just a boolean

src/pages/dashboard/instances/InstancesPage/InstancesPage.tsx
```diff
@@ -46,10 +46,10 @@ const InstancesPage: FC = () => {
         <InstancesContainer
           instanceCount={instancesCount}
           instances={instances}
-          isGettingInstances={isGettingInstances}
           selectedInstances={selectedInstances}
           setSelectedInstances={setSelectedInstances}
           onChangeFilter={clearSelection}
+          isGettingInstances={isGettingInstances}
```
gesquivelgaghi, line 52: just a general recommendation, when you undo any changes you made in a PR, its usually best to try to put things back to the way they were. its not really a big deal at all and you dont need to change it in this case, but it changes the git-blame which, like we saw today, makes it a little harder to find the source of a code decision.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.test.tsx
```diff
@@ -0,0 +1,34 @@
...
+    );
+  });
+
+  it("does not render register link", () => {
+    renderWithProviders(<InstancesEmptyState />);
+
+    expect(
+      screen.queryByRole("link", { name: "Register a new instance" }),
+    ).not.toBeInTheDocument();
+  });
```
gesquivelgaghi: you dont really have to test this since its not something that ever shows in the UI. Usually we test that something isnt there when the code branches off into a scenario where it's rendered and a scenario where its not, so we wanna make sure the branching logic is correct.

In this case, there's no branching at all, so this isnt testing anything

src/pages/dashboard/instances/InstancesPage/InstancesPage.test.tsx
```diff
@@ -1,9 +1,63 @@
...
+  useGetInstances,
+  InstancesPageActions: () => <div data-testid="instances-page-actions" />,
+}));
+
+vi.mock("../InstancesContainer", () => ({
+  default: (props: unknown) => {
+    instancesContainerSpy(props);
+    return <div data-testid="instances-container" />;
+  },
+}));
```
gesquivelgaghi: as a codebase standard, we usually avoid mocks wherever possible and treat our unit tests a little more like integration tests

src/pages/dashboard/instances/InstancesPage/InstancesPage.test.tsx
```diff
@@ -1,9 +1,63 @@
...
+    expect(screen.getByTestId("instances-page-actions")).toBeInTheDocument();
+    expect(screen.getByTestId("instances-container")).toBeInTheDocument();
+  });
+
+  it("passes count and loading state to InstancesContainer", () => {
+    useGetInstances.mockReturnValue({
+      instances: [],
+      instancesCount: undefined,
+      isGettingInstances: true,
+    });
```
gesquivelgaghi:
```suggestion
    setEnpointStatus({ path: "computers", status: "loading")
```
though you might have to add some endpointStatus handling to `src/tests/handlers/instances` if it doesnt have any. if this is the case feel free to reach out to me and I can explain it to you

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.test.tsx
```diff
@@ -170,4 +175,73 @@ describe("InstancesContainer", () => {
...
+    renderWithProviders(
+      <InstancesContainer
+        {...props}
+        instanceCount={undefined}
+        isGettingInstances
+        setSelectedInstances={() => undefined}
+      />,
+    );
+
+    expect(screen.getByRole("status")).toHaveTextContent("Loading...");
```
gesquivelgaghi:
```suggestion
    expect(screen.getByRole("status")).toBeInTheDocument();
```

its a nitpick so not really significant and its fine to leave as is, but in general you dont really need to test the  internal implementation of the loading component itself

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.test.tsx
```diff
@@ -105,18 +105,23 @@ describe("InstancesContainer", () => {
         screen.getByRole("columnheader", { name: label }),
       ).toBeInTheDocument();
 
+      const [columnCheckbox] = screen.getAllByLabelText(checkboxLabel) as [
+        HTMLElement,
+        ...HTMLElement[],
+      ];
```
gesquivelgaghi:
```suggestion
      const checkbox = screen.getByRole("checkbox", { name: checkboxLabel });
```

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.tsx
RyanLoi98: fair enough, I guess without the additional isLoading helper we won't need it anymore.

src/pages/dashboard/instances/InstancesPage/InstancesPage.tsx
RyanLoi98: Noted! Thanks for the advice!

src/pages/dashboard/instances/InstancesPage/InstancesPage.test.tsx
RyanLoi98: ok understood, I looked at how we are writing tests and I will implement something similar with the proper technique. Thanks for the feedback.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,18 @@
...
+      link={{
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/manage-computers/",
+        text: "How to manage instances in Landscape",
+      }}
+    />
+  );
+};
+
+export default InstanceEmptyState;
```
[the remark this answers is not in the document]
RyanLoi98: fixed.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,18 @@
+import type { FC } from "react";
+import EmptyState from "@/components/layout/EmptyState";
+
+const InstanceEmptyState: FC = () => {
```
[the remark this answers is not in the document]
RyanLoi98: fixed

src/pages/dashboard/instances/InstancesContainer/helpers.ts
```diff
@@ -0,0 +1,14 @@
...
+  pageSize: number,
+  instanceCount: number | undefined,
+  isFilteringInstances: boolean,
+  isGettingInstances: boolean,
+) =>
+  currentPage === 1 &&
+  pageSize === DEFAULT_PAGE_SIZE &&
+  !isGettingInstances &&
+  instanceCount === 0 &&
+  !isFilteringInstances;
```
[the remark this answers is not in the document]
RyanLoi98: removed helper function, sorry I missed that.

RyanLoi98:
I also performed this task within this PR: https://warthogs.atlassian.net/browse/LNDENG-4765



The reason why was because I made a mistake when pushing a new branch and accidentally pushed the changes to this branch. Since changes were minor and also done in this branch/task (2 lines of changes and I also updated the url in the empty instances state) it should be ok.

.changeset/petite-steaks-check.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": patch
+---
+
+Added empty state for instances
```
gesquivelgaghi:
```suggestion
Added empty state for instances and fixed broken link in scripts page
```

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.tsx
```diff
@@ -26,6 +32,40 @@ const InstancesContainer = memo(function InstancesContainer({
...
+    availabilityZones.length > 0 ||
+    tags.length > 0 ||
+    wsl.length > 0;
+
+  const hasNoInstances =
+    !isGettingInstances &&
+    currentPage === 1 &&
+    pageSize === DEFAULT_PAGE_SIZE &&
+    instanceCount === 0 &&
+    !isFilteringInstances;
```
gesquivelgaghi:
```suggestion
    !isGettingInstances &&
    instanceCount === 0 &&
    !isFilteringInstances;
```
the page size doesnt make a difference on whether there are no instances or not. If someone has 0 instances and the pageSize is 50 we still want to show them the emptyState. 

as for the currentPage, it doesnt matter because we do dynamic validation of the values so that the only value allowed for any page > 1 is only the maximum number based on the count of  instances, so we'll never run into the issue of "the user is looking for page 3 but they only have 2 pages worth of content". We just filter out the `page=3` param if thats the case.

gesquivelgaghi (commented):
Great work, just 2 minor things

.changeset/petite-steaks-check.md
RyanLoi98: Great catch, I completely forgot to update the changeset after accidentally pushing the scripts change to the wrong branch.

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.tsx
RyanLoi98: Yup I agree, I was basing if off of how we do things in scripts without thinking too deeply so great catch. I will talk to Rubin about all of this and refactor everything on that side when I get a chance.

marqode:
Looks good at first glance, once you resolve conflicts I'll review everything and approve.

RyanLoi98:
> Looks good at first glance, once you resolve conflicts I'll review everything and approve.

Awesome thanks Marc! I just made the changes so we should be good now.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,18 @@
...
+import EmptyState from "@/components/layout/EmptyState";
+
+const InstancesEmptyState: FC = () => {
+  return (
+    <EmptyState
+      title="No instances found"
+      icon="connected"
+      body="You don't have any instances registered to Landscape yet."
+      link={{
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/manage-computers/",
```
marqode: What's the reasoning for using this link? It brings me to the [classic web portal](https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/) section of the docs- I would've thought [this page](https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/web-portal-24-04-or-later/manage-instances/) would be a better fit (and it uses the term "instances" instead of "computers" as an added bonus)

Whichever link we use

src/features/scripts/components/ScriptsEmptyState/ScriptsEmptyState.tsx
```diff
@@ -25,7 +25,7 @@ const ScriptsEmptyState: FC = () => {
       icon="connected"
       body="You haven’t added any scripts yet."
       link={{
-        href: "https://ubuntu.com/landscape/docs/managing-computers",
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/manage-computers/",
```
marqode: Similar to above, it'd be good to confirm with @YanisaHS (if you haven't already) what link to use here. If it uses the same link as `InstancesEmptyState` refactor it into an export from `instances/constants.ts`

marqode (changes requested):
Everything looks good except (possibly) the documentation link. Once you've confirmed the link (ie whether it should use `/classic-web-portal/`), if both empty states use the same link it'd be best to export from a `constants.ts` so that it's easier to update later on.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
RyanLoi98: Excellent question, and the reason why I used this link is because instances can only be registered on the old portal. The new portal only discusses how to manage instance (install snaps, updates, etc), search for instances, delete them, etc. But none of that is possible during the instances empty state, so you'd need to register an instance first and to do that you need the old portal.

src/features/scripts/components/ScriptsEmptyState/ScriptsEmptyState.tsx
RyanLoi98: Again excellent point and thanks for bringing it up. For scripts the reason why I used the same link is because it was the one used before, and the link just broke because the documentation was updated. So I fixed that with the new url.


But I will ask Yanisa what she recommends we use, and then update the links accordingly and if they are the same I will make a constants file.

Thanks for your feedback and excellent catch!

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.tsx
```diff
@@ -0,0 +1,18 @@
...
+const InstancesEmptyState: FC = () => {
+  return (
+    <EmptyState
+      title="No instances found"
+      icon="connected"
+      body="You don't have any instances registered to Landscape yet."
+      link={{
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/landscape-installation-and-set-up/configure-landscape-client/",
+        text: "How to manage instances in Landscape",
+      }}
```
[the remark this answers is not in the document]
RyanLoi98: You need to install the client onto each instances before managing them, and you don't configure landscape client you install it on an instance you want to manage.. So I think my text is appropriate.

src/features/instances/components/InstancesEmptyState/InstancesEmptyState.test.tsx
```diff
@@ -0,0 +1,32 @@
...
+    renderWithProviders(<InstancesEmptyState />);
+
+    const docsLink = screen.getByRole("link", {
+      name: "How to manage instances in Landscape",
+    });
+
+    expect(docsLink).toHaveAttribute(
+      "href",
+      "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/classic-web-portal/manage-computers/",
+    );
```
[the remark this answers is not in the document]
RyanLoi98: Whoops Im making the changes right now.

src/features/scripts/components/ScriptsEmptyState/ScriptsEmptyState.tsx
```diff
@@ -25,7 +25,7 @@ const ScriptsEmptyState: FC = () => {
       icon="connected"
       body="You haven’t added any scripts yet."
       link={{
-        href: "https://ubuntu.com/landscape/docs/managing-computers",
+        href: "https://documentation.ubuntu.com/landscape/how-to-guides/web-portal/web-portal-24-04-or-later/use-remote-script-execution/",
         text: "How to manage instances in Landscape",
       }}
```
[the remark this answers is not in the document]
RyanLoi98: Yup making changes right now - my mistake.

marqode (approved):
LGTM!

src/pages/dashboard/instances/InstancesContainer/InstancesContainer.tsx
```diff
@@ -26,6 +31,34 @@ const InstancesContainer = memo(function InstancesContainer({
...
+
+  const isFilteringInstances =
+    !!query ||
+    !!status ||
+    !!os ||
+    !!contractExpiryDays ||
+    accessGroups.length > 0 ||
+    availabilityZones.length > 0 ||
+    tags.length > 0 ||
+    wsl.length > 0;
```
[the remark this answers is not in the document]
RyanLoi98: fixed!

gesquivelgaghi (approved):

PR #701, opened 2026-07-06
Fix: Autoinstall files sidepanel styling

gesquivelgaghi:
## Summary

- made the button group collapsible 
- made the title inline
- fixed & added tests

Before vs After:
image-20260605-193402 image 

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/src/features/autoinstall-files/components/AutoinstallFileDetails/AutoinstallFileDetails.tsx
+++ b/src/features/autoinstall-files/components/AutoinstallFileDetails/AutoinstallFileDetails.tsx
@@ -1,3 +1,4 @@
+import { ResponsiveButtons } from "@/components/ui";
 import { ActionButton, Button, Icon, ICONS } from "@canonical/react-components";
 import type { FC } from "react";
 import { useBoolean } from "usehooks-ts";
@@ -33,21 +34,21 @@ const AutoinstallFileDetails: FC<AutoinstallFileDetailsProps> = ({
 
   return (
     <>
-      <div className="p-segmented-control">
-        <div className="p-segmented-control__list">
+      <ResponsiveButtons
+        collapseFrom="xs"
+        buttons={[
           <Button
+            key="edit"
             type="button"
-            className="p-segmented-control__button"
             onClick={openAutoinstallFileEditForm}
             hasIcon
           >
             <Icon name="edit" />
             <span>Edit</span>
-          </Button>
-
+          </Button>,
           <ActionButton
+            key="set-as-default"
             type="button"
-            className="p-segmented-control__button"
             onClick={setAutoinstallFileAsDefault}
             disabled={autoinstallFile.is_default}
             hasIcon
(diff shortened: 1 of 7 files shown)


RyanLoi98 (approved):
LGTM. 


This wasn't part of your changes, but I noticed there were some responsiveness issues with the remove modal clipping the side panel, if you also wanted to tackle them. If not let me know and I could make a ticket for it!

Issues were from: 620 px to 1150 px. Issues stop after that.


Image

Image

Image

gesquivelgaghi:
Good catch, thanks for flagging that!

This is a known issue we have with modals opened from side panels that are missing the `renderInPortal` attribute. We've already fixed many, but there's probably multiple left still. We should make a ticket to address all the remaining ones in a targeted PR.

RyanLoi98:
> Good catch, thanks for flagging that!
> 
> This is a known issue we have with modals opened from side panels that are missing the `renderInPortal` attribute. We've already fixed many, but there's probably multiple left still. We should make a ticket to address all the remaining ones in a targeted PR.

Yup no worries! Do you want to open the ticket or should I? 

(I just don't know which other modals have issues)

gesquivelgaghi:
i made it here https://warthogs.atlassian.net/browse/LNDENG-4787

src/features/autoinstall-files/components/AutoinstallFileDetails/AutoinstallFileDetails.tsx
```diff
@@ -33,42 +35,42 @@ const AutoinstallFileDetails: FC<AutoinstallFileDetailsProps> = ({
 
   return (
     <>
-      <div className="p-segmented-control">
-        <div className="p-segmented-control__list">
+      <ResponsiveButtons
+        className={classes.marginBottom}
+        collapseFrom="xs"
```
marqode, line 40: IIRC we only support screen sizes as small as `md` or `lg` (small-ish tablet, 768 or 1036px). I see `xs` being used for a few similar cases so it may better to keep that pattern for this PR, just curious if there's some guidance I'm missing on setting breakpoints?

src/features/autoinstall-files/components/AutoinstallFileSidePanelTitle/AutoinstallFileSidePanelTitle.tsx
```diff
@@ -15,11 +15,13 @@ const AutoinstallFileSidePanelTitle: FC<AutoinstallFileSidePanelTitleProps> = ({
   version = file.version,
 }) => {
   return (
-    <div className={classes.container}>
-      {!!title && `${title} `}
-      {file.filename}, v{version}
+    <span>
+      <span className={classes.text}>
+        {!!title && `${title} `}
+        {file.filename}, v{version}
```
marqode, line 21: Should `version` prop be required? Seems like it's a required field on the `autoinstallFile` type and this would break if it's undefined

marqode (approved):
Nit comment on breakpoint, seems like `version` should be required in title component but that isn't really the focus of this PR.

src/features/autoinstall-files/components/AutoinstallFileSidePanelTitle/AutoinstallFileSidePanelTitle.tsx
gesquivelgaghi: it wont be undefined because when its not passed here it defaults to `file.version`. The prop is only there for the usage of opening the version history

src/features/autoinstall-files/components/AutoinstallFileDetails/AutoinstallFileDetails.tsx
gesquivelgaghi: not sure why we would only support down to `md` breakpoints, i haven't heard anything about it. I just thought the more we can support the better when it comes to smaller screens. But afaik, only a couple of dropdowns break at 520px and a few other things break in the 400s pixels range, but most things work well enough even there

PR #702, opened 2026-07-06
Fix: disable autofill suggestions for sensitive fields

gesquivelgaghi:
## Summary

fixes https://warthogs.atlassian.net/browse/LNDENG-4646

Will follow WCAG guidance by only disabling them for non-password fields, as explained in [this thread](https://chat.canonical.com/canonical/pl/4yz9gy94c7d7tk8kf7m3quscjr).

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
+++ b/src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
@@ -1,9 +1,3 @@
-.typeFieldsIndent {
-  border-left: 1px solid var(--vf-color-border-default, #cdcdcd);
-  margin-top: 0.5rem;
-  padding-left: 1rem;
-}
-
 .maskedInput {
   /* Mask the entered bearer token without type="password", which browsers
    * autofill (and offer to generate passwords for) even with autocomplete="off". */
--- a/src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
+++ b/src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
@@ -1,5 +1,0 @@
-.maskedInput {
-  /* Mask the entered swift password without type="password", which browsers
-   * autofill (and offer to generate passwords for) even with autocomplete="off". */
-  -webkit-text-security: disc;
-}
--- a/src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
+++ b/src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
@@ -1,5 +1,5 @@
 .maskedInput {
-  /* Mask the entered bearer token without type="password", which browsers
+  /* Mask the entered swift password without type="password", which browsers
    * autofill (and offer to generate passwords for) even with autocomplete="off". */
   -webkit-text-security: disc;
 }
--- a/src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
+++ b/src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
@@ -19,6 +19,7 @@ import {
   VALIDATION_SCHEMA,
 } from "../../constants";
 import type { EditTargetFormValues } from "../../constants";
+import classes from "./EditTargetForm.module.scss";
 
 export { TARGET_TYPE_LABELS } from "../../constants";
 
@@ -203,6 +204,7 @@ const EditTargetForm: FC<EditTargetFormProps> = ({ target }) => {
           <Input
             type="text"
             label="AWS access key ID"
+            autoComplete="off"
             help="Leave blank to keep current value"
             error={getFormikError(formik, "awsAccessKeyId")}
             {...formik.getFieldProps("awsAccessKeyId")}
@@ -210,6 +212,7 @@ const EditTargetForm: FC<EditTargetFormProps> = ({ target }) => {
           <Input
             type="text"
             label="AWS secret access key"
+            autoComplete="off"
             help="Leave blank to keep current value"
             error={getFormikError(formik, "awsSecretAccessKey")}
             {...formik.getFieldProps("awsSecretAccessKey")}
@@ -275,8 +278,10 @@ const EditTargetForm: FC<EditTargetFormProps> = ({ target }) => {
             {...formik.getFieldProps("swiftUsername")}
           />
           <Input
-            type="password"
+            className={classes.maskedInput}
+            type="text"
             label="Password"
+            autoComplete="off"
             help="Leave blank to keep current value"
             error={getFormikError(formik, "swiftPassword")}
             {...formik.getFieldProps("swiftPassword")}
--- a/src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
+++ b/src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
@@ -19,7 +19,6 @@ import {
   VALIDATION_SCHEMA,
 } from "../../constants";
 import type { EditTargetFormValues } from "../../constants";
-import classes from "./EditTargetForm.module.scss";
 
 export { TARGET_TYPE_LABELS } from "../../constants";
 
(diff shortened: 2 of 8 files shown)


src/features/publication-targets/components/TargetTypeFields/TargetTypeFields.module.scss
```diff
@@ -3,3 +3,9 @@
   margin-top: 0.5rem;
   padding-left: 1rem;
 }
+
+.maskedInput {
+  /* Mask the entered bearer token without type="password", which browsers
+   * autofill (and offer to generate passwords for) even with autocomplete="off". */
+  -webkit-text-security: disc;
+}
```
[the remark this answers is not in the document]
gesquivelgaghi: but the stylelint check passes

src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
```diff
@@ -0,0 +1,5 @@
+.maskedInput {
+  /* Mask the entered bearer token without type="password", which browsers
+   * autofill (and offer to generate passwords for) even with autocomplete="off". */
+  -webkit-text-security: disc;
+}
```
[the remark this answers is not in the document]
gesquivelgaghi: but the stylelint check passes

src/features/publication-targets/components/TargetTypeFields/SwiftFields.tsx
```diff
@@ -27,9 +27,10 @@ const SwiftFields: FC<SwiftFieldsProps> = ({ formik }) => (
       {...formik.getFieldProps("swift.username")}
     />
     <Input
-      type="password"
+      className={styles.maskedInput}
+      type="text"
       label="Password"
-      autoComplete="new-password"
+      autoComplete="off"
       required
```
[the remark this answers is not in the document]
gesquivelgaghi: we dont want the browser to suggest to create a new password. Besides, despite being non-stardard, all major browsers support -webkit-text-security

src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
```diff
@@ -275,8 +278,10 @@ const EditTargetForm: FC<EditTargetFormProps> = ({ target }) => {
             {...formik.getFieldProps("swiftUsername")}
           />
           <Input
-            type="password"
+            className={classes.maskedInput}
+            type="text"
             label="Password"
+            autoComplete="off"
             help="Leave blank to keep current value"
```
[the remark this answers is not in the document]
gesquivelgaghi: we dont want the browser to suggest to create a new password. Besides, despite being non-stardard, all major browsers support -webkit-text-security

src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
```diff
@@ -0,0 +1,5 @@
+.maskedInput {
+  /* Mask the entered swift password without type="password", which browsers
+   * autofill (and offer to generate passwords for) even with autocomplete="off". */
+  -webkit-text-security: disc;
+}
```
marqode: Hiding the text with SCSS is more fragile than using the HTML password type (see PR comments).

If we keep this, consider moving to `styles/partials/form.scss` to avoid duplication, unless we have a precedent of duplicating class names across local stylesheets.

src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
```diff
@@ -19,6 +19,7 @@ import {
   VALIDATION_SCHEMA,
 } from "../../constants";
 import type { EditTargetFormValues } from "../../constants";
+import classes from "./EditTargetForm.module.scss";
 
 export { TARGET_TYPE_LABELS } from "../../constants";
```
marqode, line 24:
```suggestion
const IGNORE_EXTENSIONS = {
  "data-bwignore": "true",
  "data-1p-ignore": "true",
  "data-lpignore": "true",
  "data-dashlane-ignore": "true",
};
```

src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
```diff
@@ -275,8 +278,10 @@ const EditTargetForm: FC<EditTargetFormProps> = ({ target }) => {
             {...formik.getFieldProps("swiftUsername")}
           />
           <Input
-            type="password"
+            className={classes.maskedInput}
+            type="text"
             label="Password"
+            autoComplete="off"
```
marqode: When testing on Chrome I didn't see any difference whether or not `autoComplete="off"` was included: setting `type="text"` blocks chrome's native autocomplete, but with `type="password"` it autocompletes regardless of whether this is present or not.

If we switch back to `type="password"` for security reasons, we can use `autocomplete="one-time-code"` to suppress native autocomplete, and the `IGNORE_EXTENSIONS` attributes to suppress password managers:


```suggestion
            autoComplete="one-time-code"
            {...IGNORE_EXTENSIONS}
```

marqode (changes requested):
Some security issues with the `type="text"` + CSS class approach for hiding the password with no autocomplete:
- screen readers will see `type="text"` and activate Character Echo, reading each character out loud as the user types them
- password will be exposed if the CSS fails to load
- browser may cache the password locally

Of these the screen reader issue feels the most pressing. I had success setting `autocomplete="one-time-code"` and adding attributes with a `{...IGNORE_EXTENSIONS}` variable to tell password manager extensions to ignore the fields.

It's possible the benefits of using `type="text"` outweigh these concerns, just want to be cautious as we try to make these forms more accessible that we don't also introduce security risks.

src/features/publication-targets/components/EditTargetForm/EditTargetForm.module.scss
gesquivelgaghi: we have that precedent bc of our css modules approach, ethan said we should avoid using global classes because it makes it harder to figure out where its being used.

on one hand i get it, on the other hand i think llms make it quite easy to find usages across the codebase, so im fine either way

src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
gesquivelgaghi: i tested in firefox and `autocomplete=off` did make a difference, but I'll test out your suggestion, since it would solve the issues you mentioned

gesquivelgaghi: firefox just ignores everything and still suggests password generation 
image

update: i also tested chromium and weirdly it also doesnt supress password managers to me:
image

src/features/publication-targets/components/EditTargetForm/EditTargetForm.tsx
```diff
@@ -203,13 +203,15 @@ const EditTargetForm: FC<EditTargetFormProps> = ({ target }) => {
           <Input
             type="text"
             label="AWS access key ID"
+            autoComplete="off"
             help="Leave blank to keep current value"
             error={getFormikError(formik, "awsAccessKeyId")}
             {...formik.getFieldProps("awsAccessKeyId")}
           />
           <Input
             type="text"
```
marqode:
```suggestion
            type="password"
```
This one seems to fall under [secrets management guidelines](https://cheatsheetseries.owasp.org/cheatsheets/Secrets_Management_Cheat_Sheet.html) and should be treated like other password fields.

src/features/publication-targets/components/TargetTypeFields/S3Fields.tsx
```diff
@@ -35,13 +35,15 @@ const S3Fields: FC<S3FieldsProps> = ({ formik }) => (
       type="text"
       label="AWS access key ID"
       required
+      autoComplete="off"
       error={getFormikError(formik, ["s3", "awsAccessKeyId"])}
       {...formik.getFieldProps("s3.awsAccessKeyId")}
     />
     <Input
       type="text"
```
marqode:
```suggestion
      type="password"
```
same as above

marqode (changes requested):
I like this approach but I think AWS secret key should be treated the same as other passwords.

RyanLoi98 (approved):
LGTM!

marqode (approved):
LGTM!

PR #705, opened 2026-07-09
feat: adopt Pragma design system (global styles and tokens)

rubinaga:
## Summary

Introduces [Pragma](https://github.com/canonical/pragma), Canonical's next-generation
design system and the successor to vanilla-framework and `@canonical/react-components`,
into landscape-ui. This first step wires Pragma's global styles and design tokens in
app-wide, without migrating any components yet.

- **Dependencies:** add `@canonical/react-ds-global@0.29.1` and `@canonical/styles@0.29.0`
  (pinned exactly; Pragma is pre-1.0).
- **Global styles:** import `@canonical/styles` in `src/main.tsx` after the vanilla SCSS
  bundle. Loading last, Pragma's typography and reset apply the app-wide type scale on
  shared element selectors (`h1`–`h6`, `p`, `body`, `a`, form controls). Verified in the
  production bundle that Pragma's rules order after vanilla and win on shared selectors.
- **Type resolution:** declare the CSS-only `@canonical/styles` module in
  `src/vite-env.d.ts` so the side-effect import type-checks.
- **Test config:** `vitest.config.ts` inlines `@canonical/react-ds-global`
  (`test.server.deps.inline`) so its side-effect `.css` imports resolve through the existing
  `styleMock` alias, and strips the package's dangling sourcemap references — required for
  any future test that renders a Pragma component.
- **Docs:** add a "Pragma design system" section to `docs/FRONTEND.md`.

Note: this is a global-styles change — Pragma's type scale now applies across the whole app,
including pages not yet migrated to Pragma components. This is intentional. Component
migration, forms, and icon delivery are out of scope and handled in follow-ups.

## Release Impact

Change type (tick one):

- [x] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

### DIFF
--- a/.changeset/shiny-tables-ring.md
+++ b/.changeset/shiny-tables-ring.md
@@ -1,0 +1,5 @@
+---
+"landscape-ui": patch
+---
+
+Add Pragma design system styles
--- a/.changeset/shiny-tables-ring.md
+++ b/.changeset/shiny-tables-ring.md
@@ -1,5 +1,2 @@
 ---
-"landscape-ui": patch
 ---
-
-Add Pragma design system styles
--- a/docs/FRONTEND.md
+++ b/docs/FRONTEND.md
@@ -145,30 +145,28 @@ Because the current repo still contains some domain hooks under `src/hooks/`, tr
 
 ## Pragma design system (@canonical/pragma)
 
-Pragma (`@canonical/react-ds-global` + `@canonical/styles`) is Canonical's next-generation
-design system and the successor to vanilla-framework and `@canonical/react-components`. Adoption
-is incremental: Pragma is installed and its global styles are active, and components move to it
-over time. Surfaces not yet migrated keep using vanilla / `@canonical/react-components`.
+Pragma is Canonical's next-generation design system and the successor to vanilla-framework and
+`@canonical/react-components`. The app currently has Pragma's A1 setup in place through
+`@canonical/styles`, `@canonical/react-ds-global`, `@canonical/react-ds-global-form`, and
+`@canonical/ds-assets`; component migration remains incremental, so many surfaces still use vanilla
+and `@canonical/react-components`.
 
-**Using Pragma components**
+`src/main.tsx` loads the Vanilla SCSS bundle first, then `@canonical/styles`, then
+`@canonical/react-ds-global-form/dist/esm/index.css`. `@canonical/styles` brings in the active
+global design tokens, reset, and typography rules; the form stylesheet provides package-level CSS
+variables for future Pragma form components. These are Vite side-effect imports rather than SCSS
+`@import`s because the package CSS entrypoints resolve correctly from the Vite entrypoint.
 
-- Import components from `@canonical/react-ds-global` and style them with Pragma's own design
-  tokens.
-- Pragma versions are pinned exactly; it is pre-1.0, so upgrade deliberately.
+Font faces still come from vanilla-framework in this codebase, so there is no separate
+`@canonical/styles/fonts` import at the moment. Pragma icon assets come from
+`@canonical/ds-assets/icons` and are served at `/icons` by `createPragmaIconsPlugin` in
+`vite.config.ts`; production builds copy the same assets to `dist/icons`, and Debian packaging
+includes that directory in the deployed dashboard files.
 
-**Global styles**
-
-- `@canonical/styles` is imported once in `src/main.tsx`, after the vanilla SCSS bundle. Loading
-  last, Pragma's typography and reset apply the app-wide type scale on shared element selectors
-  (`h1`–`h6`, `p`, `body`, `a`, form controls), so import order matters.
-- The CSS-only `@canonical/styles` module is declared in `src/vite-env.d.ts` so the import
-  type-checks.
-
-**Testing Pragma components**
-
-- `vitest.config.ts` inlines `@canonical/react-ds-global` (`test.server.deps.inline`) so its
-  side-effect `.css` imports resolve through the existing `styleMock` alias, and strips the
-  package's dangling sourcemap references. Tests that render a Pragma component depend on this.
+When migrating components, import non-form components from `@canonical/react-ds-global` and form
+components from `@canonical/react-ds-global-form`. `vitest.config.ts` currently inlines
+`@canonical/react-ds-global` so its side-effect CSS imports are handled by the existing
+`styleMock` alias and its dangling sourcemap references are stripped during tests.
 
 ## Testing
 
--- a/docs/FRONTEND.md
+++ b/docs/FRONTEND.md
@@ -143,6 +143,33 @@ Because the current repo still contains some domain hooks under `src/hooks/`, tr
 - The build runs `tcm` against `src/**` for SCSS modules, so CSS module type generation must continue to work.
 - Stylelint covers SCSS files in CI, and Prettier checks Markdown and other formatted assets.
 
+## Pragma design system (@canonical/pragma)
+
+Pragma (`@canonical/react-ds-global` + `@canonical/styles`) is Canonical's next-generation
+design system and the successor to vanilla-framework and `@canonical/react-components`. Adoption
+is incremental: Pragma is installed and its global styles are active, and components move to it
+over time. Surfaces not yet migrated keep using vanilla / `@canonical/react-components`.
+
+**Using Pragma components**
+
+- Import components from `@canonical/react-ds-global` and style them with Pragma's own design
+  tokens.
+- Pragma versions are pinned exactly; it is pre-1.0, so upgrade deliberately.
+
+**Global styles**
+
+- `@canonical/styles` is imported once in `src/main.tsx`, after the vanilla SCSS bundle. Loading
+  last, Pragma's typography and reset apply the app-wide type scale on shared element selectors
+  (`h1`–`h6`, `p`, `body`, `a`, form controls), so import order matters.
+- The CSS-only `@canonical/styles` module is declared in `src/vite-env.d.ts` so the import
+  type-checks.
+
+**Testing Pragma components**
+
+- `vitest.config.ts` inlines `@canonical/react-ds-global` (`test.server.deps.inline`) so its
+  side-effect `.css` imports resolve through the existing `styleMock` alias, and strips the
+  package's dangling sourcemap references. Tests that render a Pragma component depend on this.
+
 ## Testing
 
 - Follow [testing/index.md](testing/index.md) for the repository testing contract and [verification/index.md](verification/index.md) for completion criteria.
--- a/package.json
+++ b/package.json
@@ -26,9 +26,11 @@
     "tcm:watch": "pnpm tcm:run --watch"
   },
   "dependencies": {
+    "@canonical/ds-assets": "0.29.0",
     "@canonical/landscape-openapi": "0.0.70",
     "@canonical/react-components": "4.6.0",
     "@canonical/react-ds-global": "0.29.1",
+    "@canonical/react-ds-global-form": "0.29.1",
     "@canonical/styles": "0.29.0",
     "@monaco-editor/react": "4.7.0",
     "@sentry/react": "10.56.0",
--- a/package.json
+++ b/package.json
@@ -26,12 +26,12 @@
     "tcm:watch": "pnpm tcm:run --watch"
   },
   "dependencies": {
-    "@canonical/ds-assets": "0.29.0",
+    "@canonical/ds-assets": "0.31.0",
     "@canonical/landscape-openapi": "0.0.70",
     "@canonical/react-components": "4.6.0",
-    "@canonical/react-ds-global": "0.29.1",
-    "@canonical/react-ds-global-form": "0.29.1",
-    "@canonical/styles": "0.29.0",
+    "@canonical/react-ds-global": "0.31.0",
+    "@canonical/react-ds-global-form": "0.31.0",
+    "@canonical/styles": "0.31.0",
     "@monaco-editor/react": "4.7.0",
     "@sentry/react": "10.56.0",
     "@tanstack/react-query": "5.101.0",
--- a/package.json
+++ b/package.json
@@ -28,6 +28,8 @@
   "dependencies": {
     "@canonical/landscape-openapi": "0.0.70",
     "@canonical/react-components": "4.6.0",
+    "@canonical/react-ds-global": "0.29.1",
+    "@canonical/styles": "0.29.0",
     "@monaco-editor/react": "4.7.0",
     "@sentry/react": "10.56.0",
     "@tanstack/react-query": "5.101.0",
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -1,4 +1,5 @@
 import "./styles/index.scss";
+import "@canonical/styles";
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import * as Sentry from "@sentry/react";
--- a/src/main.tsx
+++ b/src/main.tsx
@@ -1,5 +1,6 @@
 import "./styles/index.scss";
 import "@canonical/styles";
+import "@canonical/react-ds-global-form/dist/esm/index.css";
 import { StrictMode } from "react";
 import { createRoot } from "react-dom/client";
 import * as Sentry from "@sentry/react";
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -1,7 +1,76 @@
 import fs from "fs";
+import { createRequire } from "module";
 import * as path from "path";
 import { defineConfig, loadEnv } from "vite";
 import react from "@vitejs/plugin-react";
+import type { Plugin, PreviewServer, ViteDevServer } from "vite";
+
+const require = createRequire(import.meta.url);
+const PRAGMA_ICONS_ROUTE = "/icons/";
+const PRAGMA_ICONS_DIR = path.join(
+  path.dirname(require.resolve("@canonical/ds-assets/package.json")),
+  "icons",
+);
+
+const getPragmaIconPath = (requestUrl: string | undefined) => {
+  const pathname = new URL(requestUrl ?? "", "http://localhost").pathname;
+
+  if (!pathname.startsWith(PRAGMA_ICONS_ROUTE)) {
+    return null;
+  }
+
+  const iconPath = path.normalize(
+    decodeURIComponent(pathname.slice(PRAGMA_ICONS_ROUTE.length)),
+  );
+
+  if (!iconPath || iconPath.startsWith("..") || path.isAbsolute(iconPath)) {
+    return null;
+  }
+
+  return path.join(PRAGMA_ICONS_DIR, iconPath);
+};
+
+const servePragmaIcons = (server: ViteDevServer | PreviewServer) => {
+  server.middlewares.use((req, res, next) => {
+    const filePath = getPragmaIconPath(req.url);
+
+    if (!filePath) {
+      next();
+      return;
+    }
+
+    fs.stat(filePath, (error, stats) => {
+      if (error || !stats.isFile()) {
+        next();
+        return;
+      }
+
+      res.setHeader("Content-Type", "image/svg+xml");
+      fs.createReadStream(filePath).pipe(res);
+    });
+  });
+};
+
+export const createPragmaIconsPlugin = (): Plugin => {
+  let root = process.cwd();
+  let outDir = "dist";
+
+  return {
+    name: "serve-pragma-icons",
+    configResolved(config) {
+      root = config.root;
+      outDir = config.build.outDir;
+    },
+    configureServer: servePragmaIcons,
+    configurePreviewServer: servePragmaIcons,
+    writeBundle() {
+      fs.cpSync(PRAGMA_ICONS_DIR, path.resolve(root, outDir, "icons"), {
+        recursive: true,
+        force: true,
+      });
+    },
+  };
+};
 
 export default defineConfig(({ mode }) => {
   const env = loadEnv(mode, process.cwd(), "");
@@ -9,6 +78,7 @@ export default defineConfig(({ mode }) => {
   return {
     plugins: [
       react(),
+      createPragmaIconsPlugin(),
       {
         name: "exclude-msw",
         apply: "build",
--- a/vite.config.ts
+++ b/vite.config.ts
@@ -19,9 +19,21 @@ const getPragmaIconPath = (requestUrl: string | undefined) => {
     return null;
   }
 
-  const iconPath = path.normalize(
-    decodeURIComponent(pathname.slice(PRAGMA_ICONS_ROUTE.length)),
-  );
+  let decodedIconPath: string;
+
+  try {
+    decodedIconPath = decodeURIComponent(
+      pathname.slice(PRAGMA_ICONS_ROUTE.length),
+    );
+  } catch (error) {
+    if (error instanceof URIError) {
+      return null;
+    }
+
+    throw error;
+  }
+
+  const iconPath = path.normalize(decodedIconPath);
 
   if (!iconPath || iconPath.startsWith("..") || path.isAbsolute(iconPath)) {
     return null;
--- a/vitest.config.ts
+++ b/vitest.config.ts
@@ -86,11 +86,8 @@ export default defineConfig({
   },
   plugins: [
     {
-      // @canonical/react-ds-global ships .js.map files whose "sources" point at
-      // src/ paths it doesn't publish. Inlining it for tests (server.deps.inline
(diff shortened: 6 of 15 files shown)


package.json
```diff
@@ -26,6 +26,8 @@
   "dependencies": {
     "@canonical/landscape-openapi": "0.0.70",
     "@canonical/react-components": "4.6.0",
+    "@canonical/react-ds-global": "0.29.1",
+    "@canonical/styles": "0.29.0",
     "@monaco-editor/react": "4.7.0",
```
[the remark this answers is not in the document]
rubinaga: it will be part of A2 integration, seems unnecessary to remove it from here

.changeset/shiny-tables-ring.md
```diff
@@ -0,0 +1,5 @@
+---
+"landscape-ui": patch
+---
+
+Add Pragma design system styles
```
RyanLoi98: I might be wrong in this particular case, but I thought non UI/user facing changes don't get changeset files - and technically this isn't UI/user facing yet.

src/main.tsx
```diff
@@ -1,4 +1,5 @@
 import "./styles/index.scss";
+import "@canonical/styles";
```
RyanLoi98, line 2: Just something I noticed when reading the [Pragma Docs](https://docs.google.com/document/d/1WhO8JYCEFpa6v-rXthe19qPmXvAAm1PU8zkKJoVnRIw/edit?tab=t.0) — they mention adding this import as well: @import "@canonical/react-ds-global-form/dist/esm/index.css";. I noticed we don't have it right now. Not sure if we need this import right now or if we will import them later on. Just wanted to point it out.

RyanLoi98 (dismissed):
LGTM!

Just one thing I had to mention was if we needed that additional import from the [Pragma documentation](https://docs.google.com/document/d/1WhO8JYCEFpa6v-rXthe19qPmXvAAm1PU8zkKJoVnRIw/edit?tab=t.0) or not (I tried to import it but it caused some  errors and broke the build so im guessing it is not needed, and copilot is also saying we dont need it for our particular use).



I also tested the Pragma accordion component and it seemed to load and work just fine so it should be good.


Something to note was that  dark mode styling for the Pragma accordion was really black compared to how our dark mode is now so not sure if that was a bug or a design feature, because this is the type of dark mode that hurts the eyes when reading for a long period of time.

Edit- I checked the [Pragma story book](https://ds-react-global.canonical.com/?path=/docs/documentation-introduction--docs) and the dark mode there had that shade of black as well, so it doesn't appear to be a bug with how we installed/imported Pragma. So everything seems to be good on our end!!!



Nice work Rubin!

vite.config.ts
```diff
@@ -1,14 +1,84 @@
...
+const getPragmaIconPath = (requestUrl: string | undefined) => {
+  const pathname = new URL(requestUrl ?? "", "http://localhost").pathname;
+
+  if (!pathname.startsWith(PRAGMA_ICONS_ROUTE)) {
+    return null;
+  }
+
+  const iconPath = path.normalize(
+    decodeURIComponent(pathname.slice(PRAGMA_ICONS_ROUTE.length)),
+  );
```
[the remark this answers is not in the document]
rubinaga: Added

src/main.tsx
rubinaga: I was going to add it down the line when working on forms, but might as well add it here. It's changed now

.changeset/shiny-tables-ring.md
rubinaga: I changed it back. I thought it maybe made sense to say that we are changing design system styles in our changelog but maybe not

RyanLoi98: Yeah I'm not 100% sure either. But my thought was because no visual ui changes are made as of yet, it is better to not talk about it right now to avoid confusion. This is more of an internal matter. But someone else can provide their opinion

.changeset/shiny-tables-ring.md
```diff
@@ -0,0 +1,2 @@
+---
+---
```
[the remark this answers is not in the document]
RyanLoi98: An empty changeset file should be ok for non UI changes, my thought was because no visual ui changes are made as of yet, it is better to not talk about it right now to avoid confusion. This is more of an internal matter since we are not using Pragma yet - more so preparing for it.

docs/FRONTEND.md
```diff
@@ -143,6 +143,31 @@ Because the current repo still contains some domain hooks under `src/hooks/`, tr
...
+
+`src/main.tsx` loads the Vanilla SCSS bundle first, then `@canonical/styles`, then
+`@canonical/react-ds-global-form/dist/esm/index.css`. `@canonical/styles` brings in the active
+global design tokens, reset, and typography rules; the form stylesheet provides package-level CSS
+variables for future Pragma form components. These are Vite side-effect imports rather than SCSS
+`@import`s because the package CSS entrypoints resolve correctly from the Vite entrypoint.
+
+Font faces still come from vanilla-framework in this codebase, so there is no separate
+`@canonical/styles/fonts` import at the moment. Pragma icon assets come from
+`@canonical/ds-assets/icons` and are served at `/icons` by `createPragmaIconsPlugin` in
```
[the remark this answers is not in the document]
RyanLoi98: I guess we can add the trailing slash to make it clearer, but honestly it is so minor that I don't think it is an issue if we do not change it.

RyanLoi98 (approved):
Reviewed the code and tested it and it with some buttons and LGTM!


GJ Rubin!

ethanashaw (dismissed):

vitest.config.ts
```diff
@@ -68,6 +83,40 @@ export default defineConfig({
       "@": resolve(__dirname, "src"),
     },
   },
+  plugins: [
+    {
+      // @canonical/react-ds-global ships .js.map files whose "sources" point at
+      // src/ paths it doesn't publish. Inlining it for tests (server.deps.inline
+      // above) makes Vite read those maps and warn "points to missing source
+      // files". Strip the dangling sourcemap reference and return an empty map so
+      // Vite never looks for the missing sources. Keeps test output pristine.
```
marqode: comment could be simplified to something like:

```
// @canonical/react-ds-global's sourcemaps point to unpublished src/ paths.
// Strip them and return an empty map to silence Vite's "missing source" test warnings.
```

marqode (commented):
I noticed some table fields (like Date on /publications and /local-repositories) wrap now where before they didn't. Not sure if the fix belongs in this PR or if we want a separate `feature/pragma-style-fixes` to build off of this (same goes for comment below).

Image

Also this might be intentional, but the vertical spacing of nav items is a lot tighter with pragma than without. Maybe we should confirm the change with Dani to see if we need a custom style to keep the UI consistent.

without pragma:
Image

with pragma:

vitest.config.ts
RyanLoi98: Good catch mark, I didn't notice styling change while I was testing it. It must be doing something in the background to interfere with Vanilla.

gesquivelgaghi:
> I noticed some table fields (like Date on /publications and /local-repositories) wrap now where before they didn't. Not sure if the fix belongs in this PR or if we want a separate `feature/pragma-style-fixes` to build off of this (same goes for comment below).

Weird, I'm not getting this at all

> Also this might be intentional, but the vertical spacing of nav items is a lot tighter with pragma than without. Maybe we should confirm the change with Dani to see if we need a custom style to keep the UI consistent.

This actually seems to be a lot bigger in scope than just navigation items. Looks like the default line height changed from `24px` to `18.4px` which is causing the spacing to be smaller all over. Should specify that to Dani when discussing this. 

But there are a couple other issues in the navigation related to this, which can be seen in marc's screenshot above:

- the nav items with a chevron are now thicker than the ones without a chevron, which means non-uniform spacing
- the external link icon is misaligned with the text now

.changeset/shiny-tables-ring.md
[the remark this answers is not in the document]
gesquivelgaghi: Agreed, especially because the user does not know or care about pragma

gesquivelgaghi (changes requested):
i think this should be fixed before its merged, since its a regression 

Image

RyanLoi98 (changes requested):
I agree with Gabe and Marc, it seems like there are really small visual bugs introduced with Pragma (great catch by them), and I think we need to talk to Dani about the next steps before we add Pragma, and we need to address these bugs too.


Otherwise it worked perfectly fine, and I could use the Pragma components, so it was a great bit of effort Rubin - good job!

ethanashaw:
Aren't these issues going to be addressed later? Or did I misunderstand the PR description?

gesquivelgaghi:
> Aren't these issues going to be addressed later? Or did I misunderstand the PR description?

What I understood from the description is that the pragma components will be imported in future PRs, but my suggestion is not to import the pragma components now, simply to add a small fix to the global styling of the external icons so that we dont merge a visual regression right before the break & the point release. If necessary, the fix can be removed when we add the pragma icons.

My concern is that if we leave it as is, there's a chance the follow-up pragma icons PR fixing the external icon alignment issue doesnt get merged until after the point release

vite.config.ts
```diff
@@ -1,14 +1,96 @@
...
+
+const require = createRequire(import.meta.url);
+const PRAGMA_ICONS_ROUTE = "/icons/";
+const PRAGMA_ICONS_DIR = path.join(
+  path.dirname(require.resolve("@canonical/ds-assets/package.json")),
+  "icons",
+);
+
+const getPragmaIconPath = (requestUrl: string | undefined) => {
+  const pathname = new URL(requestUrl ?? "", "http://localhost").pathname;
```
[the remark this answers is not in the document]
marqode: This feels like overkill since we call `decodeURICOmponent() later on`, but I would add destructuring to avoid eslint warnings:
`const { pathname } = new URL(requestUrl ?? "", "http://localhost");`

Same for lines 73-74

marqode (commented):
Looking good overall! Mostly minor suggestions, though I wonder if the font change issue I'm seeing (screenshots below) should be treated as a potential regression
 
- default line height still changes from ~35px to 32px. Feel free to disregard if Dani already approved this
- PR description versioning out of date- still has `@0.29.1` but package.json has been bumped to `@0.31.0` for all dependencies.
- update PR title with `fix: ` to match Change type and empty changeset
- this might be a local caching issue, but I don't see the Ubuntu font everywhere- instead a lot of places show "Family name: .SF NS" as the rendered font

in this branch:
Image

vs in main: 
Image

src/styles/partials/utils.scss
```diff
@@ -1,4 +1,10 @@
 @import "vanilla-framework/scss/settings_font";
+@import "vanilla-framework/scss/settings_spacing";
+
+// Vanilla's paragraph bottom margin; Pragma zeroes element spacing by design.
+.u-margin--bottom {
+  margin-bottom: 1.125rem;
+}
```
ethanashaw, line 7: Would it be reasonable for to use the `p` selector instead so we don't have to apply this class everywhere?

ethanashaw (changes requested):
I had a look through, only found a few things I could find:

---

I'm seeing a different font on notifications:
Image

---

There are a few Vite build warnings if you could look into that:
```
/icons/information.svg#information referenced in /icons/information.svg didn't resolve at build time, it will remain unchanged to be resolved at runtime

/icons/success.svg#success referenced in /icons/success.svg didn't resolve at build time, it will remain unchanged to be resolved at runtime

/icons/warning.svg#warning referenced in /icons/warning.svg didn't resolve at build time, it will remain unchanged to be resolved at runtime

/icons/error.svg#error referenced in /icons/error.svg didn't resolve at build time, it will remain unchanged to be resolved at runtime
```

src/styles/partials/utils.scss
rubinaga: I thought of this approach but it felt counter-intuitive to how pragma wants to approach `<p>` tags and tyopgraphy (with less spacing). We already have cases of using `u-no-margin--bottom` in <p> tags. 

I haven't addressed those in this PR because it seems out of scope, and since `u-no-margin--bottom` adds slightly negative margin to `<p>` tags I didn't change it for now, but it's something to change in the future.

ethanashaw (approved):

yurii-vasyliev (dismissed):
Reviewed the full diff, built both this branch and `main`, and compared the compiled CSS cascade. `tsc` and `vite build` are clean with no warnings, CI is green, and the branch merges onto `main` without conflicts. Nice work on the shim file: its claims about `@canonical/styles` declaring layers but importing normalize unlayered check out in the built bundle.

A few things I'd like addressed before merging, ranked by importance.

### 1. Code font regression (blocking)

normalize's `code,kbd,samp { font-family: monospace }` and `pre { font-family: monospace }` have the same specificity as Vanilla's `code,kbd,pre,samp { font-family: Ubuntu Mono variable, ... }` rule and load after it, so they win. Pragma's typography only restyles the `.code` *class*, not the element, so nothing restores it.

Affected: every `CodeSnippet` (script code, activity details, logs, export details, error fallback), the bare `<code>` in `SearchHelpPopup` and `MirrorFilterHelpButton`, and the `<kbd>` hints in `SearchBoxWithSavedSearches`.

Fix is one more rule in `src/styles/partials/pragma-cascade-shims.scss`, same `:not([hidden])` pattern as the existing ones, restoring `$font-monospace` on `code, kbd, pre, samp`.

### 2. Fieldset padding

Same mechanism: normalize's `fieldset { padding: .35em .75em .625em }` beats Vanilla's `fieldset { padding: calc(1rem - 1.5px) }`. Only two fieldsets exist (`CheckboxGroup` and `OTPInput`). Please check the attach page OTP input visually and add it to the shim file if it moved.

### 3. The 126 `u-margin--bottom` additions are a maintenance trap

Every future `<p>` in the app will now render with zero bottom margin unless the author remembers to add the class, and there is no lint to catch it. A scoped restore rule in the shim file (one line, same `:not([hidden])` trick, `p:not([hidden]) { margin-bottom: $sp-unit * 1.5 }` or equivalent) would cover all current and future paragraphs, including the ones rendered inside `@canonical/react-components`, and the shim file already documents that it gets deleted once components migrate.

The comment in the shim file argues the class is "permanent by design", but that design only holds once Pragma components own those surfaces. Until then the app is still Vanilla-styled and should get Vanilla's paragraph spacing by default. Happy to discuss with Dani if you see it differently, but I'd rather not ship a footgun for every future PR.

### 4. Dependency footprint (upstream issue, not blocking)

`@canonical/react-ds-global@0.34.0` lists `@canonical/storybook-config` as a *runtime* dependency, which drags storybook, svelte and lit into every install. The lockfile grows from 809 to 996 packages. Nothing reaches the bundle, but this is an upstream packaging bug worth reporting to the Pragma team, and possibly excluding via a `pnpm.overrides` entry in the meantime.

### 5. Bundle and deploy size (for awareness)

| | `main` | this branch |
|---|---|---|
| `index-*.css` raw | 649 KB | 796 KB |
| `index-*.css` gzip | 87 KB | 101 KB |
| woff2 emitted | 204 KB | 1.4 MB |
| `dist/icons` | – | 664 KB |

Both the Ubuntu and Ubuntu Sans families are emitted; only Ubuntu Sans should actually get fetched since `body` now overrides `html`'s font-family. Worth confirming in the network tab.

### 6. PR description and changeset

- The description is stale: it lists 2 of the 4 new packages, and says nothing about the icons pipeline (`vite.config.ts`, `debian/rules`, the release workflow), the shim file, or the ~100 component edits.
- The PR silently deletes three unused components and their tests (`ProfileAssociationInfo`, `RepositoryProfileFormAptSourcesPanel`, `RepositoryProfileFormTabs`). They are dead code so I'm fine with removing them, but it should be called out in the description.
- The changeset is empty. A change that swaps the app font family and type scale deserves a CHANGELOG line.

### 7. Minor (from the open Copilot threads)

- `servePragmaIcons` pipes a read stream with no `error` handler. Dev/preview only, but a one-liner.
- `cleanUrl(id).replace(/^\/@fs/, "")` in `vitest.config.ts` leaves a leading slash on Windows paths. Not a problem on Linux CI, mentioning for completeness.

yurii-vasyliev (commented):
Follow-up to my review above. I rebuilt this branch and `main` and diffed the compiled `index-*.css` selector by selector, rather than spot-checking. That turned up more of the same class of problem, plus two corrections to what I wrote.

### Corrections to my own review

- It's **146** `u-margin--bottom` additions across **91** files, not 126.
- My suggested `$sp-unit * 1.5` is wrong. `$sp-unit-ratio` is `0.5`, so `$sp-unit` is `0.5rem` and that expression gives `0.75rem`. Vanilla's actual paragraph margin is `1.125rem`. Use the literal, or `$sp-unit * 2.25`.

### 1. Headings regress exactly like `<p>`, and are mostly unpatched

Pragma sets `--space-after: 0` on `h1`–`h6` as well as `p`, and then applies `h1,h2,h3,h4,h5,h6,p,.p { margin-block-end: calc(var(--space-after,0) * var(--baseline-height)) }`. The only thing that restores heading spacing is `.editorial h1,.editorial h2 { --space-after: 2 }` — scoped to a class this app never uses.

Headings carrying a `p-heading--N` class are safe: that's `(0,1,0)` and out-specifies Pragma's element rule. Bare headings are not. This PR added `u-margin--bottom` to two `<h5>`s, which suggests the symptom was noticed but not traced to the mechanism.

Live example in a file this PR already edits — `CronHelp.tsx:41` and `:80`:

```tsx
<h3 className="p-text--small-caps">Example</h3>
<h3 className="p-text--small-caps">Accepted values</h3>
```

`.p-text--small-caps` only sets `font-weight: 550`, so these `<h3>`s take their margin from the bare `h3` selector, which Pragma now zeroes. Both headings collapse onto the content below. The `<p>` two lines down in the same component *was* patched.

So the shim I proposed should cover `h1, h2, h3, h4, h5, h6, p`, not just `p`.

### 2. Three more unshimmed normalize wins

Same mechanism as the code font and fieldset issues, found by diffing every element selector:

- `hr { height: 0 }` beats Vanilla's `.p-rule,…,hr { height: 1px }`. No bare `<hr>` in `src/`, but worth confirming `@canonical/react-components` doesn't emit any.
- `b,strong { font-weight: bolder }` beats Vanilla's `550`. `bolder` is relative, so **59 usages** in `src/` now render at 700.
- `small { font-size: 80% }` beats Vanilla's `.875rem`. 12 usages.

Also a gap in the existing shim: it restores `line-height` and `margin-bottom` for `select`/`textarea`, but the *same* normalize rule (`button,input,optgroup,select,textarea`) also sets `font-family: inherit` and `font-size: 100%`, which beat Vanilla's explicit `font-size: 1rem` / Ubuntu family on those two elements. `[type='text']` and friends keep Vanilla's values because they're attribute selectors. Net effect: `<select>` and `<textarea>` drift away from every other field in the same form.

### 3. `form.scss` is a behaviour change, not a shim

```scss
.p-form-help-text,
.p-form-validation__message {
  margin-bottom: 1.125rem;
}
```

Both of these are *classes*, `(0,1,0)`. Pragma's element rules never touched them — on `main` they already get `margin-bottom: .95rem` from Vanilla. The new rule has equal specificity and loads later, so it wins and adds `0.175rem` to every help text and validation message in the app.

That makes the shim file's note that these are "permanent, not shims" rest on a false premise. Either drop this rule, or set it to `.95rem` if the intent was to hold spacing steady.

### 4. The icons pipeline is unused, and will 404 in the deb

I grepped the emitted bundle: **zero** `/icons/` references in the built JS or CSS. Nothing imports `@canonical/react-ds-global`, and `@canonical/react-ds-global-form`'s CSS doesn't reference icons either. So the Vite plugin, `rolldownOptions.external`, 664 KB of SVGs, `debian/rules`, `debian-build.sh` and the release workflow change all ship with no consumer.

More importantly, the route is root-absolute. `release-and-build.yml:141` builds with `VITE_ROOT_PATH=/new_dashboard/`, and `debian/install` maps `dist` to `opt/canonical/landscape/dashboard`. Pragma emits `url("/icons/chevron-down.svg#chevron-down")`, which resolves against the server root, not the dashboard base — so it 404s in the packaged build the moment A2 lands and starts rendering Pragma components. Dev and preview work because `VITE_ROOT_PATH` is `/` there, so this won't show up locally.

Worth either splitting the icons work into the PR that actually needs it, or making the route base-aware now while it's cheap.

### 5. Two open Copilot threads are wrong and can be resolved

- The "this changeset is empty, the Changesets workflow will fail" comment is incorrect. `validate.yml:226` says explicitly: *"Empty changesets count — they still create a file."* The `Changeset` check is passing.
- The "`main.test.tsx` will hit Unknown file extension .css unless `@canonical/styles` is inlined" comment is also incorrect. I ran it on this branch: 7/7 pass. The bare-specifier CSS entrypoints resolve fine; no extra `server.deps.inline` entry is needed.

On the changeset itself — I'll defer to the earlier thread with @RyanLoi98 and @gesquivelgaghi, but flag that the scope has moved since that discussion. It was settled when this was "no visual change yet"; it now swaps the app-wide font family and type scale and touches 91 files of markup. That feels like it crossed the line into something a user would notice.

rubinaga:
@yurii-vasyliev thanks for the review. I’ve addressed the accidental regressions caused by Pragma importing Normalize outside its declared layer:

- Restored the existing code font for bare code, kbd, pre, and samp.
- Restored Vanilla’s b/strong weight and bare small size.
- Restored the affected select/textarea styles.
- Restored fieldset padding while preserving the intentional OTP and CheckboxGroup overrides.
- Removed the unnecessary form help/validation margin override.
- Added icon stream error handling and corrected Vite /@fs/ path handling.
- Made icon paths follow VITE_ROOT_PATH, so development uses /icons/ and the packaged dashboard uses /new_dashboard/icons/.

I’m keeping explicit paragraph and heading spacing rather than adding a global restore shim. The Pragma team confirmed that removing element-owned spacing is intentional and that applications should add spacing where their layouts require it. A global p/heading rule would reverse intentional Pragma behavior and could also affect future A2/A3 components. The existing utility additions preserve today’s layouts while following that direction.

The PR description has been updated to reflect the complete dependency set, icon pipeline, cascade compatibility work, explicit spacing, and dead-code removal.

src/styles/partials/pragma-cascade-shims.scss
```diff
@@ -0,0 +1,54 @@
+// Temporary shims for Pragma's unlayered cascade.
```
marqode, line 1: nit: shims not pinned to a specific pragma version, could cause some UI drift once pragma updates

src/styles/partials/base.scss
```diff
@@ -29,6 +29,11 @@ body.is-dark {
   --vf-color-link-visited: #{$colors--dark-theme--focus};
 }
 
+.p-notification__title {
```
marqode, line 34: nit: add comment like "Pragma typography resets these; restore Vanilla defaults under shim"

vite.config.ts
```diff
@@ -1,7 +1,134 @@
...
+      res.setHeader("Content-Type", "image/svg+xml");
+      const iconStream = fs.createReadStream(filePath);
+      iconStream.on("error", (streamError) => {
+        if (res.headersSent) {
+          res.destroy(streamError);
+          return;
+        }
+
+        next(streamError);
+      });
```
marqode, line 85: When headersSent is true, we destroy the response pipe with no trace. In dev this makes truncated/silent-failed SVGs hard to debug. 

Suggest adding a log line (e.g. console.error("Failed to serve icon", filePath, streamError) or server.config.logger?.warn(...)) so this doesn't look like a network blip locally.

vitest.config.ts
```diff
@@ -69,6 +86,37 @@ export default defineConfig({
...
+    {
+      // @canonical/react-ds-global's sourcemaps point to unpublished src/ paths.
+      // Strip them and return an empty map to silence Vite's "missing source" test warnings.
+      name: "strip-pragma-sourcemaps",
+      enforce: "pre",
+      load(id) {
+        const filePath = viteIdToFilePath(id);
+        if (
+          !filePath.includes("@canonical/react-ds-global/") ||
+          !filePath.endsWith(".js")
```
marqode: Nit: could filter on `.mjs` here as well, eg `!/\.m?jsx?$/.test(filePath)`

marqode (approved):
LGTM, a few nits in comments. 

One coverage gap that could be worth closing is that currently createPragmaIconsPlugin / getPragmaIconPath have no unit tests. Could add a test file for the new path-traversal guards (getPragmaIconPath rejects .. and absolute paths after decodeURIComponent), fs.stat error handling, and writeBundle directory copy.

marqode (approved):
LGTM! Nice work on this one.

ethanashaw (approved):

PR #691, opened 2026-06-29
feat: Add LRO information to Locals

gesquivelgaghi:
## Summary

Summarize the changes made in this pull request. Include any relevant context or background information that would help reviewers understand the purpose and scope of the changes.

## Release Impact

Pick the target branch (see [RELEASES.md](../RELEASES.md) for the full model):

- [ ] `dev` — internal testing → publishes to `ppa-build-dev`
- [x] `main` — next beta → publishes to `ppa-build`
- [ ] `release/YY.MM` — point release on a maintained cycle → publishes to `ppa-build-YY.MM` (and to `ppa-build-stable` if this is the currently-promoted branch). Specify cycle: `release/____`

Change type (tick one):

- [ ] Patch (fix)
- [x] Minor (feature)
- [ ] Major (breaking)

> The change-type label is informational and only affects how the entry is rendered in the CHANGELOG. The actual version is computed from the branch and CalVer cycle — `pnpm changeset`'s `patch`/`minor` choice does not influence it.

## Checklist

- [x] **Changeset added** — I have run `pnpm changeset` (or `pnpm changeset --empty` if no CHANGELOG line is warranted) and committed the resulting `.md` file. Required for PRs targeting `main` and `release/*`; enforced by the `Changeset check` workflow.
- [ ] **UI verified** — I have verified the changes locally.
- [x] **Linting clean** — No linting errors are present (especially in `scripts/`).

## Versioning Reminder

> [!IMPORTANT]
> Landscape UI uses **CalVer** (`YY.MM.Point.Build`). Version derivation by branch:
>
> - `main` / `point/*` → `YY.MM.0.<run>-beta`
> - `dev` → `YY.MM.0.<run>-dev`
> - `release/YY.MM` → `YY.MM.1.<run>` (cycle pinned by branch name)

PR #708, opened 2026-07-13
feat: resolve API coverage routes from contracts instead of heuristics

yurii-vasyliev:
## Summary

Follow-up to the API usage mapping POC (#703) — targets `feat/map-api-usage`, not `main`. This replaces the heuristic `normalizeRoute` with contract-backed route resolution and makes the contract layer pluggable, so the real BE contract can take over as endpoints migrate.

## Why

Running the POC over the full suite showed the regex heuristics misclassify routes in both directions, on the exact traffic this repo produces:

| Actual URL | Heuristic normalized it to | Problem |
|---|---|---|
| `/api/v2/auth/supported-providers` | `/api/v2/auth/:id` | Static endpoint erased (any hyphenated segment collapsed) |
| `/api/v2/auth/oidc-providers/1` | `/api/v2/auth/:id/1` | Inverted: static segment became `:id`, the real id stayed literal |
| `/api/v2/repository/ubuntu-archive-info` | `/api/v2/repository/:id` | Static endpoint erased by the `dynamicParents` list |
| `/publications/{uuid}:publish` | `/publications//:id:publish` | Double slash; custom verb handling inconsistent |

For a tool whose purpose is gap detection, false merges are worse than no normalization — untested endpoints get reported as covered. Meanwhile the exact patterns already exist: every MSW handler declares one, and the Go BE ships a real OpenAPI document in `@canonical/landscape-openapi`.

## Design

The recorder stays raw (unchanged, apart from dumping a handler-pattern manifest during runs). Route resolution happens at aggregation time via **contract connectors** consulted in priority order:

1. **`openapi`** — the Go BE's published contract, mounted under `/debarchive`. Authoritative where it exists.
2. **`v1-actions`** — the legacy Python RPC API (`/api/?action=...`); route identity is the `action` param, universe scanned from `isAction()` dispatches in the handlers.
3. **`msw-handlers`** — declared MSW patterns; for the unspecced Python v2 API these are the best machine-readable contract that exists today.
4. **`heuristic`** — last resort, and everything it emits is labeled a guess.

Every route carries `backend: "v1" | "v2" | "go"` and `source` provenance. Structurally identical declarations across connectors (OpenAPI `{name}` vs MSW `{mirrorId}`) dedupe, higher priority wins. When a Python endpoint migrates to Go, the new spec version simply out-prioritizes the mock — no config change.

`contract-coverage/migration-map.ts` is a hand-maintained equivalence table across generations; the report shows live per-layer traffic for each entry (e.g. mirror sync: v1 declared but unexercised, go exercised), and ids are validated against the declared universe to catch typos.

## Results (full suite, 3182 tests)

```
10482 interactions -> 202 routes exercised, 240 declared, 38 unexercised, 0 drift
v1: 50 exercised / 62 declared
v2: 126 exercised / 147 declared
go: 26 exercised / 31 declared
```

Zero drift: every observation resolved to a declared contract; the heuristic never fired. The provenance labels also immediately surfaced that `POST /debarchive/v1beta1/operations:batchGet` exists only in the mocks — the operations service is missing from the shipped Go spec.

## Notes

- `src/tests/coverage/` was not usable as a directory name — `.gitignore`'s `coverage` rule (vitest output) swallows it; hence `contract-coverage`.
- The report gained sections: `unexercised` (declared, zero traffic), `drift` (traffic no contract claims), `migration`, `warnings`.
- Usage unchanged: `npx vitest run`, then `npx tsx src/tests/aggregate-coverage.ts`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/tests/aggregate-coverage.ts
+++ b/src/tests/aggregate-coverage.ts
@@ -1,195 +1,59 @@
 import fs from "fs";
-import path from "path";
-import crypto from "crypto";
+import { buildReport } from "./contract-coverage/aggregate";
+import { MIGRATION_MAP } from "./contract-coverage/migration-map";
+import { REGISTRY_PATH, REPORT_PATH } from "./contract-coverage/paths";
+import { createHeuristicSource } from "./contract-coverage/sources/heuristic";
+import { createMswHandlerSource } from "./contract-coverage/sources/msw";
+import { createOpenApiSource } from "./contract-coverage/sources/openapi";
+import { createV1ActionSource } from "./contract-coverage/sources/v1-actions";
+import type { Observation } from "./contract-coverage/types";
 
-const INPUT_PATH = path.resolve(import.meta.dirname, "msw-api-registry.jsonl");
-const OUTPUT_PATH = path.resolve(
-  import.meta.dirname,
-  "msw-contract-coverage.json",
-);
+// The Go debarchive service is mounted under this prefix in front of the
+// spec's own /v1beta1 (VITE_API_URL_DEB_ARCHIVE = /debarchive/v1beta1/).
+const DEB_ARCHIVE_MOUNT = "/debarchive";
 
-interface TrafficMetrics {
-  totalHits: number;
-  statuses: Record<number, number>;
-  // Only populated for state-mutating requests (POST, PUT, PATCH)
-  contracts?: {
-    status: number;
-    requestHash?: string;
-    responseHash?: string;
-    requestPayload: unknown;
-    responsePayload: unknown;
-  }[];
-}
-
-interface ContractCoverageReport {
-  generatedAt: string;
-  summary: {
-    totalInteractionsLogged: number;
-    uniqueRoutesFound: number;
-  };
-  routes: Record<string, Record<string, TrafficMetrics>>;
-}
-
-/**
- * Heuristically converts raw execution URLs into normalized route patterns
- * Strips query params, collapses UUIDs, hex values, and dynamic path slugs.
- */
-function normalizeRoute(rawUrl: string): string {
-  try {
-    const url = new URL(rawUrl);
-    let { pathname } = url;
-
-    // 1. Standardize trailing slashes
-    if (pathname.endsWith("/") && pathname.length > 1) {
-      pathname = pathname.slice(0, -1);
-    }
-
-    // 2. Normalize standard UUIDs -> :id
-    pathname = pathname.replace(
-      /[0-9a-f]{8}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{4}-[0-9a-f]{12}/gi,
-      "/:id",
-    );
-
-    // 3. Normalize known dynamic REST subpaths (e.g., /mirrors/third-party-mirror -> /mirrors/:id)
-    const dynamicParents = [
-      "mirrors",
-      "operations",
-      "computers",
-      "instances",
-      "users",
-      "repository",
-    ];
-    dynamicParents.forEach((parent) => {
-      const regex = new RegExp(`(?<=\\/${parent}\\/)[^\\/]+`, "g");
-      pathname = pathname.replace(regex, ":id");
-    });
-
-    // 4. Clean up trailing alphanumeric hashes/slugs with hyphens (e.g., mirror-ffff-llll-dddd)
-    pathname = pathname.replace(/\/[a-z0-9]+-[a-z0-9-]+/gi, "/:id");
-
-    return pathname;
-  } catch {
-    // Fixed: Handled potential undefined element accessor under strict noUncheckedIndexedAccess rule
-    return rawUrl.split("?")[0] ?? rawUrl;
-  }
-}
-
-/**
- * Generates a deterministic hash of payload structures to detect true mutations
- */
-// Fixed: Swapped explicit 'any' out for 'unknown'
-function hashPayload(payload: unknown): string {
-  if (!payload) return "empty";
-  const str = typeof payload === "string" ? payload : JSON.stringify(payload);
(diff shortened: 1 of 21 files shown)


src/tests/setup.ts
```diff
@@ -68,6 +68,23 @@ server.events.on("response:mocked", ({ request, response }) => {
...
+try {
+  const manifest = server.listHandlers().map((handler) => {
+    const info = handler.info as { method?: unknown; path?: unknown };
+    return {
+      method: String(info.method ?? ""),
+      path: String(info.path ?? ""),
+      isRegExpPath: info.path instanceof RegExp,
+    };
+  });
+  fs.writeFileSync(MANIFEST_PATH, JSON.stringify(manifest, null, 2));
```
marqode, line 85:
```suggestion
try {
  fs.writeFileSync(MANIFEST_PATH, JSON.stringify(manifest, null, 2));
} catch (error) {
  throw new Error(`[FATAL] MSW manifest write failed at ${MANIFEST_PATH}: ${error}`);
}
```
Nit: fail hard if manifest write fails

src/tests/aggregate-coverage.ts
```diff
@@ -1,195 +1,59 @@
...
+import { createHeuristicSource } from "./contract-coverage/sources/heuristic";
+import { createMswHandlerSource } from "./contract-coverage/sources/msw";
+import { createOpenApiSource } from "./contract-coverage/sources/openapi";
+import { createV1ActionSource } from "./contract-coverage/sources/v1-actions";
+import type { Observation } from "./contract-coverage/types";
+
+// The Go debarchive service is mounted under this prefix in front of the
+// spec's own /v1beta1 (VITE_API_URL_DEB_ARCHIVE = /debarchive/v1beta1/).
+const DEB_ARCHIVE_MOUNT = "/debarchive";
+
```
marqode, line 18:
```suggestion
if (!fs.existsSync(MANIFEST_PATH)) {
  console.error(`[ERROR] MSW manifest missing. Run: pnpm vitest`);
  process.exit(1);
}
```
Similar to `REGISTRY_PATH` guard below. Or the two could be combined since the intent is similar.

marqode (approved):
LGTM! Thanks for catching the regex issue in the PoC, this feels like a solid improvement. A few ideas for follow-up, could be added here or in another PR:
- add unit tests for pattern matching, connectors (URL resolution, file parsing) aggregator, and migration-map validation
- More of a question, but I remember [some discussion](https://github.com/canonical/landscape-ui/pull/691#discussion_r3515927788) around the proto not fully typing LROs; should we account for that in `openapi.ts`?
- add scripts to package.json:
```
"coverage-report": "tsx src/tests/aggregate-coverage.ts",
"coverage:full": "pnpm run coverage && pnpm run coverage-report"
```

src/tests/setup.ts
yurii-vasyliev: Applied in d6a388a6a — the manifest write now throws a `[FATAL]` error so a broken manifest fails the run instead of silently skewing the report.

src/tests/aggregate-coverage.ts
yurii-vasyliev: Applied in d6a388a6a — went with the combined form: one guard checks both generated inputs (registry + manifest) and exits with a "run the test suite first" hint. The soft fallback inside the msw connector became dead code with this, so it's removed.

yurii-vasyliev:
Thanks for the review! All three suggestions applied in d6a388a6a: the fail-hard manifest write, the combined input guard, and the `coverage-report` / `coverage:full` scripts (with `tsx` and `yaml` added as devDependencies so they're self-contained).

On the LRO question — good timing. Syncing to the current `@canonical/landscape-openapi` (0.0.70) turned out to answer it partially:

- The package now ships `openapi.yaml` (gnostic-generated) instead of `openapi.json`; the connector loads either.
- The new spec uses explicit path params (`/v1beta1/mirrors/{mirror}`) instead of the old regex wildcards (`{name_1}` + `pattern: mirrors/[^/]+`), and it **now types `GET /v1beta1/operations/{operation}`** — so single-operation polling is contract-backed.
- What's still missing from the spec is `POST /v1beta1/operations:batchGet`, which the frontend exercises (the report attributes it to `msw-handlers` instead of `openapi` — that provenance label is exactly the mock-vs-contract drift signal). So I'd say the connector needs no special LRO handling for *route* coverage; where the untyped-LRO discussion from #691 will bite is a future payload-schema validation layer, since `Operation.response`/`metadata` are `Any`-typed. Worth keeping the batchGet gap on the radar for the Go team though.

One bonus finding: the 0.0.59→0.0.70 param rename (`{name}` → `{mirror}`) was caught by the migration-map typo guard, which is the maintenance behavior we wanted from the hand-maintained table.

Unit tests for the matcher/connectors/aggregator: agreed, planning those as a follow-up commit on this PR.

src/tests/contract-coverage/paths.ts
```diff
@@ -0,0 +1,19 @@
+import path from "path";
+
+const TESTS_DIR = path.resolve(import.meta.dirname, "..");
+
```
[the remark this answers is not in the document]
yurii-vasyliev: `import.meta.dirname` is standard Node.js API (stable since 20.11), and this repo pins `engines.node >= 24` — so it's guaranteed available here. Empirically verified on this head: the full suite (3,222 tests, every file importing `paths.ts` via `setup.ts`) is green, and the aggregator CLI runs under tsx; if `dirname` were `undefined`, `path.resolve` would throw immediately in all of those. The parent POC branch already uses `import.meta.dirname` in `aggregate-coverage.ts` as well. Keeping as-is; `import.meta.url` + `fileURLToPath` would only matter for Node < 20.11 portability, which isn't a constraint.

PR #723, opened 2026-07-28
Fix/batchget partial success

marqode:
## Summary

This PR improves partial-failure handling for batch lookups used by Publications and related views, and updates UI behavior to clearly surface missing sources. Resolves [LNDENG-4812](https://warthogs.atlassian.net/browse/LNDENG-4812)

Operations handler fix was merged as part of #726 scoped for pre-point release.

### What changed

1. Added partial-success support to batch query hooks for:
   - mirrors
   - local repositories
   - operations
   - publication targets

2. Updated MSW handlers and test fixtures to return realistic partial responses, including unreachable resources per batch request.

3. Propagated unreachable state into publications UI so missing sources are handled explicitly:
   - non-existent mirror/local sources are no longer rendered as links
   - affected rows now show a warning icon with the label Source not found
   - applied consistently in:
     - Publications list
     - Associated publications list
     - Publication details / side panel
     - Publication target details where associated publications are shown

4. Updated integration/component tests to cover:
   - partial-success/unreachable hook behavior
   - Source not found rendering and link suppression
   - operation status behavior under partial failures

### Why

Previously, missing resources could silently degrade into incomplete display mapping and still appear as normal links, which was misleading. This change uses API-provided unreachable data as the source of truth, making failures visible and preventing dead navigation paths.

### Notes

1. Notification behavior remains single-slot (latest notification overwrites previous), so multiple near-simultaneous failures may display as one visible notification. This appears to be an existing notification-system limitation, now more noticeable due to additional partial-failure reporting.

### Validation

1. Updated tests for touched publications/operations paths pass locally.
2. Mock server behavior now aligns with new hook contract for partial-success responses.
## Release Impact
Note: Final target is main, but changes from #717 are required since those are also included in this landscape-proto version.

Change type (tick one):

- [X] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Versioning Reminder

> [!IMPORTANT]
>

[LNDENG-4812]: https://warthogs.atlassian.net/browse/LNDENG-4812?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


yurii-vasyliev (changes requested):
**TL;DR:** The approach is sound and matches the ticket's acceptance criteria, but there is **one confirmed failing test on this branch**. Note that because this PR is opened against a feature branch (`fix/require-distributions-archs`) rather than `main`/`dev`, our CI (linters and tests) does not run — which is why the failing test went unnoticed.

## Confirmed bug: failing test

`OperationStatusCell.test.tsx` — "calls debug when some operations are unreachable" **fails** (verified locally in a clean checkout of this branch):

```
AssertionError: expected 'Failed to fetch operation: operations…' to contain 'operation(s)'
Received: "Failed to fetch operation: operations/non-existent"
```

Root cause: the hooks call `pluralize(unreachable.length, ["operation"])` with the default `countType: "none"`, which returns only the word — no count, and never the literal `"operation(s)"` the test expects. The same applies to the mirror/local messages ("Failed to fetch mirror: …" with no count). The likely intent was `pluralize(unreachable.length, ["operation"], "exact")` → "1 operation". Please fix the hooks to use `"exact"` and update the assertion.

Related inconsistency: `useBatchGetPublicationTargets` hand-rolls `` `${unreachable.length} publication target(s)` `` instead of using `pluralize` like its three siblings — please pick one style.

## Other findings

- **Dead code:** the four new mock exports (`batchGetLocalNamesWithMissing`, `batchGetMirrorNamesWithMissing`, `batchGetOperationNamesWithMissing`, `batchGetPublicationTargetNamesWithMissing`) are never imported anywhere. Either use them in the tests (which currently inline these names as string literals) or drop them.
- **Duplicated helper:** `isMissingSource` is copy-pasted identically in `PublicationsList.tsx` and `AssociatedPublicationsList.tsx`, and both re-hardcode the `"Mirror" | "Local repository"` strings from `getSourceType`. It belongs in `src/features/publications/helpers.ts` next to `getSourceType`, with a test.
- **Inconsistent icon styling:** `PublicationsList` uses a scss module (`.missingSource`, `.marginRight`) while `AssociatedPublicationsList` and `PublicationDetails` use a literal `{" "}` for spacing. Also `` Icon name={`${ICONS.warning} ${classes.marginRight}`} `` smuggles a CSS class through the `name` prop — it works and matches the existing pattern in `OperationStatusContent`, but `Icon` has a `className` prop for this. At minimum, please make the three new sites consistent with each other.
- **`useBatchGetOperations` doesn't expose `unreachable`** to consumers, unlike the mirror/local hooks. That's fine for the AC — a missing operation falls through to `OperationStatusContent`'s existing "Unable to determine" state — but the asymmetry in hook contracts is worth a comment, or a follow-up if LNDENG-4810 will need it.
- **Notification UX:** `debug()` calls `notify.error`, and it fires inside `queryFn`, so users get an error toast on every refetch (mount, stale refetch), and with multiple unreachable resource types only the last notification survives (acknowledged in the PR description). Worth confirming with UX that a persistent error toast per page load is intended — the inline "Source not found" state arguably makes the toast redundant for mirrors/locals.
- **Base branch / release impact:** the PR targets `fix/require-distributions-archs`, while the Release Impact checkbox says `main`. If this is intentional stacking, fine — just make sure the base PR merges first and this gets retargeted, otherwise the changeset/release checklist won't apply as described. And since CI is skipped on feature-branch PRs, please run lint + the affected tests locally before merging.

## What's good

- The hook changes are consistent and backward-compatible (return keys preserved; new `unreachable` fields additive with safe defaults).
- MSW handlers correctly gate `unreachable` on `return_partial_success`, mirroring the real API contract.
- Good test coverage for the new UI states: link suppression, tooltip text, side-panel rendering, and debug reporting are all exercised, including a realistic new mock publication pointing at a missing mirror.
- `useMemo` dependency arrays were correctly extended everywhere `unreachableSourceNames` was added — easy to miss with `react-table` column defs.

yurii-vasyliev (dismissed):
All points from my previous review have been addressed — thanks! Removing the notify/debug path from the hooks entirely was a nice resolution to both the pluralize bug and the toast-on-refetch concern, and the shared `MissingSourceLabel` + `isMissingSource` helper cleanup looks good. Full CI green after retargeting to `main`.

src/features/operations/components/OperationStatusCell/OperationStatusCell.test.tsx
```diff
@@ -28,4 +31,44 @@ describe("OperationStatusCell", () => {
 
     expect(await screen.findByText("Published")).toBeInTheDocument();
   });
+
+  it("renders status when some operations are unreachable", async () => {
+    const [, missingOperationName] = batchGetOperationNamesWithMissing;
+    if (!missingOperationName) {
+      throw new Error("Missing operation name fixture");
+    }
```
ethanashaw: Should this be an `assert`? (In other locations as well)

ethanashaw (approved):

yurii-vasyliev (approved):

PR #726, opened 2026-07-30
Fix: use partial success in batch get operations api hook

marqode:
## Summary

[Resolves LNDENG-4812](https://warthogs.atlassian.net/browse/LNDENG-4812) for only BatchGetOperations, other changes will be included in #723 once all backend changes are in.

## Release Impact

Change type (tick one):

- [X] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Reviewer Setup

**MSW (Mock Service Worker)**

- [X] MSW must be enabled (`VITE_MSW_ENABLED=true` in `.env.local`)
- [ ] MSW not required
- [X] Not Applicable

**Backend**

- [X] No specific backend branch required
- [] Not Applicable

With either MSWs enabled or the landscape-packaging backend (both repos on `main`) go to /repositories/local-repositories. Confirm that at least one operation is unreachable (shows "Unable to determine") but the rest of the operations load with other status. This means the entire batch of operations did not fail

**Screenshots** _(optional)_

image

---

## Versioning Reminder

> [!IMPORTANT]
>


yurii-vasyliev (changes requested):
## Review

### Overview

Bumps `@canonical/landscape-openapi` 0.0.78 → 0.0.80 (which adds `returnPartialSuccess`/`unreachable` to the batchGet request/response schemas) and opts `useBatchGetOperations` into partial success, so one unreachable operation no longer fails the entire batch. The existing "Unable to determine" fallback in `OperationStatusContent` handles operations missing from the lookup, so the hook change alone drives the UI in the screenshot.

### Main issue: request field name contradicts the OpenAPI contract

The hook sends `{ names, return_partial_success: true }`, but the published spec in 0.0.80 declares the JSON field as **`returnPartialSuccess`** (camelCase) — verified in both `openapi.yaml` and the generated `BatchGetOperationsRequest` type. gRPC-JSON transcoders often accept proto-style snake_case too, which is presumably why local verification passed, but:

- Every other request body in the codebase uses the camelCase JSON names (`displayName`, `preserveSignatures`, …).
- If any layer in the chain (proxy, stricter transcoder config, future codegen'd client) honors only the documented JSON name, the flag is **silently ignored** and the bug this PR fixes comes back with no error to notice.

Please change it to `returnPartialSuccess: true`.

### Related: hand-rolled types where generated ones now exist

The local `BatchGetOperationsResponse` interface predates this PR, but 0.0.80 now ships `BatchGetOperationsRequest`/`BatchGetOperationsResponse` — importing them instead of extending the local copy would have caught the field-name mismatch at compile time, and makes the newly added (currently unused) `unreachable?: string[]` declaration unnecessary to hand-maintain.

### Test coverage — the fix is effectively untested

- The MSW `operations:batchGet` handler filters mocks by `names` and returns whatever matches — it has **always behaved as partial-success** and never reproduces the all-or-nothing failure being fixed. No test can distinguish before/after this change, and nothing asserts the flag is sent.
- Suggestion, consistent with the mock-hardening in #717: make the handler mimic the real contract — error the whole batch when a name is unknown and the flag is absent; return partial results plus `unreachable` when it's set.
- Also note: **no CI checks ran on this PR** (base is a feature branch), so even the existing suite hasn't validated the branch.

### Minor

- **Changeset:** typo "operatios" → "operations"; and "Other resources with similar behavior will be updated in another PR" is process context that will render verbatim in the CHANGELOG — better kept in the PR description.
- **Stacked base:** targets `fix/require-distributions-archs` while Release Impact ticks `main`. Fine if the stacking is intentional (GitHub will retarget after #717 merges), but worth stating in the summary — which is otherwise still the unfilled template.
- Stray whitespace-only change (blank line before `return lookup;`).

### Coordination note

My local branch `fix/batchget-partial-success` covers the broader multi-resource version of this work (UI for unreachable resources, mock data/handlers, api hooks) — let's sync so the mirrors/locals/publication-targets follow-up doesn't collide or duplicate.

### Verdict

Requesting changes for the `return_partial_success` → `returnPartialSuccess` field name — it works today only by transcoder leniency and contradicts the contract this same PR upgrades to. The generated-types import and a partial-success-aware MSW handler are strongly recommended; the rest is cosmetic.

marqode:
- Updated field name to `returnPartialSuccess`
- Hardened mock handler to mimic real contract
- Update changeset and PR description
- ran tests locally; CI will run here once `fix/require-distributions-archs` is merged to `main`

yurii-vasyliev (approved):
Re-reviewed `49e2777` — the blocking issue and both strongly-recommended items are resolved:

- **Field name**: the hook now sends `returnPartialSuccess` and, better, types the body as the generated `BatchGetOperationsRequest`, so the spelling is compile-checked against the contract. Verified `0.0.80` declares `returnPartialSuccess?: boolean`.
- **MSW handler**: now mirrors the real contract — 404 (`code: 5`) when a name is unknown and the flag is absent, partial results + `unreachable` when it's set. Nice.
- **Changeset**: typo and process context gone.
- Keeping the local `BatchGetOperationsResponse` with the comment explaining the `Operation`-type divergence is a fair call.

Since CI can't run on the stacked base, I ran the checks locally: full suite **3493/3493 green** (585 files) and `tsc --noEmit` clean.

Non-blocking notes, mostly for before merge / follow-up:

- **PR description**: the Summary section is still the unfilled template, and the stacked-base note isn't there — please fill it in before merging (the comment said the description was updated, but only the "Where to Find" section was).
- **Coverage follow-up**: the hardened handler now makes an end-to-end test possible, but none exercises it — e.g. an `OperationProvider` test with `[knownName, unknownName]` asserting the known status renders and the unknown one shows "Unable to determine". As it stands, nothing fails if `returnPartialSuccess` stops being sent. Fine as a follow-up, perhaps alongside the multi-resource branch we're syncing on.
- Micro-nit: `unreachable?: string[]` in the local response interface is declared but never consumed — harmless, and presumably the follow-up will use it.

marqode:
updated PR description, coverage and micro-nit points to be covered in #723

PR #727, opened 2026-07-31
fix: add feature flag gate for localRepository.lastImportTime

marqode:
## Summary

Gates the (not-yet-backend-supported) `Last import` field behind a new `local-repository-last-import` feature flag in the local repositories list and details views.

## Changes

- Added `"local-repository-last-import"` to FeatureKey.ts.
- LocalRepositoriesList.tsx: filters out the `Last import` column when the flag is disabled (same pattern as `ScriptList`'s `associated_profiles` column).
- ViewLocalRepositoryDetailsTab.tsx: conditionally renders the `Last import` `InfoGrid.Item`.
- Added `local-repository-last-import` to the features mock (enabled by default, for MSW-driven scenarios elsewhere in the app).
- Tests in both components mock `useAuth()` directly (matching ScriptList.test.tsx) to deterministically force `isFeatureEnabled` on/off and verify the column/item is hidden when the flag is disabled, avoiding a race with the async features query that the MSW `server.use(...)` approach would have introduced.

## Testing

- `pnpm vitest` for the two affected test files (11 tests passing).

## Release Impact

Change type (tick one):

- [X] Patch (fix)
- [ ] Minor (feature)
- [ ] Major (breaking)

## Checklist

## Reviewer Setup

**MSW (Mock Service Worker)**

- [ ] MSW must be enabled (`VITE_MSW_ENABLED=true` in `.env.local`)
- [ ] MSW not required
- [ ] Not Applicable

**Backend**

- [ ] No specific backend branch required
- [ ] Not Applicable

_Describe where in the UI the changes are visible and how to navigate there (if applicable). Paste a direct URL to the new page/component if applicable (e.g. `http://localhost:5173/...`):_

> **Example:** To test restarting an instance — navigate to the home page → click **Instances** → select an instance → open the **Operations** action menu → click **Restart**.
>
> _describe here_

**Testing Instructions** _(optional)_

_Describe any special testing instructions:_

1.

**Screenshots** _(optional)_

---

## Versioning Reminder

> [!IMPORTANT]
>

### DIFF
--- a/.changeset/neat-pillows-swim.md
+++ b/.changeset/neat-pillows-swim.md
@@ -1,0 +1,5 @@
+---
+"landscape-ui": patch
+---
+
+Gate the local repository import time field behind the server-driven `local-repository-last-import` feature flag.
--- a/src/features/local-repositories/components/LocalRepositoriesList/LocalRepositoriesList.test.tsx
+++ b/src/features/local-repositories/components/LocalRepositoriesList/LocalRepositoriesList.test.tsx
@@ -1,20 +1,35 @@
-import { API_URL } from "@/constants";
+import type { AuthContextProps } from "@/context/auth";
+import useAuth from "@/hooks/useAuth";
 import { renderWithProviders } from "@/tests/render";
 import { describe, it, expect } from "vitest";
 import LocalRepositoriesList from "./LocalRepositoriesList";
 import { repositories } from "@/tests/mocks/localRepositories";
-import { features } from "@/tests/mocks/features";
-import server from "@/tests/server";
-import { generatePaginatedResponse } from "@/tests/server/handlers/_helpers";
 import { getAllByRole, screen } from "@testing-library/react";
 import { NO_DATA_TEXT } from "@/components/layout/NoData";
 import userEvent from "@testing-library/user-event";
-import { http, HttpResponse } from "msw";
+
+vi.mock("@/hooks/useAuth");
+
+const authContextValues: AuthContextProps = {
+  logout: vi.fn(),
+  authorized: true,
+  authLoading: false,
+  setUser: vi.fn(),
+  user: null,
+  redirectToExternalUrl: vi.fn(),
+  safeRedirect: vi.fn(),
+  isFeatureEnabled: () => true,
+  hasAccounts: true,
+};
 
 describe("LocalRepositoriesList", () => {
   const user = userEvent.setup();
 
-  it("renders table with column headers and pagination", async () => {
+  beforeEach(() => {
+    vi.mocked(useAuth).mockReturnValue(authContextValues);
+  });
+
+  it("renders table with column headers and pagination", () => {
     renderWithProviders(<LocalRepositoriesList repositories={repositories} />);
 
     expect(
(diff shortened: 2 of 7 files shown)


yurii-vasyliev (dismissed):
## Review

### Overview

Gates the not-yet-backend-supported "Last import" field behind the new `local-repository-last-import` feature flag in the two places it renders: the `LocalRepositoriesList` column (filtered via the new `id: "last_import"`, same pattern as `ScriptList`) and the `ViewLocalRepositoryDetailsTab` `InfoGrid.Item` (conditional render). Adds the `FeatureKey` union member, a default-enabled features-mock entry, and a hide-when-disabled test per component. All CI checks pass.

### What's solid

- **Coverage is complete**: the list column and the details item are the only two non-test render sites of `lastImportTime` on the branch, and both are gated.
- **The test approach is the right one**: switching from the MSW override to `vi.mock("@/hooks/useAuth")` with an explicit `isFeatureEnabled: () => false` resolves the false-positive risk Copilot flagged (`isFeatureEnabled` returns `false` while the features query is still loading), and matches the established pattern in the guard tests.
- Mechanical details check out: `database_key: 16` is unique in the mock, the `FeatureKey` member is alphabetically placed, and each negative assertion is anchored by a positive one (`Packages` header, `Description` item) so the queries can't pass on an empty render.

### Nits (non-blocking)

1. **PR body is stale** — it still describes the earlier revision's `server.use(...)` MSW pattern, which the current code no longer uses. Please update it; and the Reviewer Setup section is unfilled / "Linting clean" unchecked (ESLint passes, so presumably just forgotten).
2. **Mixed vitest import style** — the test files import `describe, it, expect` explicitly but use `vi`/`beforeEach` as globals. Works with `globals: true`, but add them to the import line for consistency with sibling files.
3. The whole-module auth mock in `beforeEach` applies to the files' pre-existing tests too (replacing the real auth context with `user: null` etc.). Harmless since these components only consume `isFeatureEnabled` — just noting the other tests no longer exercise the real auth path.
4. Trivial: `"last_import"` appears in both the column def and the filter — a shared const would prevent silent drift.

### Behavioral note (not a defect)

While the features query loads, `isFeatureEnabled` is `false`, so the column/item pops in after load for flag-enabled users — identical to `ScriptList`'s `associated_profiles` behavior, so consistent.

### Verdict

Approving — correct, complete, conventional, and the test approach already went through one round of hardening. The only thing I'd actually ask for is the PR body update; the rest is take-it-or-leave-it.

yurii-vasyliev (approved):
## Re-review of d866d005 — approving

All nits from my previous review are addressed:

- `beforeEach, vi` added to the vitest imports in both test files
- `LAST_IMPORT_FIELD` const shared between the column `id` and the filter
- PR body rewritten to accurately describe the `useAuth()` mock approach (including the rationale for avoiding the MSW race)

The final push just merges current `main` (with #722) into the branch; nothing in this PR's scope changed. All seven CI checks pass on the current head — the earlier Changeset failure was a `no merge base` shallow-checkout artifact right after the rebase, and the rerun cleared it.

One leftover cosmetic: the Reviewer Setup checkboxes in the body are still unticked — feel free to fill them in on the way to merge, not blocking.

PR #728, opened 2026-07-31
Feat: api contract eval agentic workflow

marqode:
Disclaimer: mostly vibe-coded, based on [this design doc](https://docs.google.com/document/d/1RpiMgBoNey1Fky0l45rUgqjjT2X5OqL6DGD3NdCuqB4/edit?tab=t.tht50f3beek6). Any contributions are welcome.

## Summary

Foundation of the **Agentic Evaluation Loop** for API contract monitoring: deterministic detection of gaps between frontend API usage (MSW contract-coverage report) and backend API-contract test coverage (`e2e/docker-stack/api/` Playwright specs), plus the LLM plumbing that will turn top gaps into draft test specs. Gap detection is 100% deterministic — the LLM only elaborates computed gaps. Zero new dependencies; provider-agnostic (GitHub Models today, OpenRouter/Gemini via env vars later).

Orchestrator, renderer, pnpm scripts, and the scheduled workflow follow in a stacked PR.

> **Base:** `feat/integration-tests` — the extraction target only exists there until it merges.

## Changes

- **`eval-loop/types.ts`** — shared interfaces + hand-rolled `assert*()` validators (field-path errors; no zod, zero new deps).
- **`eval-loop/collect-gaps.ts`** — deterministic collector: report validation with fail-fast errors; TS-AST extraction of `request.<method>(url)` calls (literals + template literals → `{param}`; dynamic URLs → warnings); gap ranking (`totalHits` desc, mutating first) with `/debarchive` mount-prefix matching; CLI with `--report/--spec-dir/--out`.
- **`eval-loop/llm-client.ts`** — zero-dep OpenAI-compatible fetch client: env factory (`LLM_API_KEY` + optional `LLM_BASE_URL`/`LLM_MODEL`, defaults GitHub Models), 60s timeout, one 429/5xx retry, rich errors, mock client; key never logged.
- **`eval-loop/prompt.ts`** — minimal prompt: top-5 gaps with observed payloads as inert JSON, one exemplar spec, strict-JSON contract, 12k-char guard.
- **`docs/testing/coverage.md`** — new eval-loop section.
- **Fix:** fixture spec moved under `fixtures/repo/e2e/` so vitest's `**/e2e/**` exclude covers it.

## Tests

- **33 new unit tests green ×3** (collect-gaps 15, llm-client 13, prompt 5); full contract-coverage suite **73/73** — no regressions.
- Failing-first proof captured per module; eslint 0 problems; prettier clean.
- **Manual QA:** fixture-driven CLI produces schema-valid `gaps.json` (mutating route ranked first); LLM client live-stubbed (request shape, redacted auth, 401/timeout/bad-body); prompt QA (4,322 chars, 5/5 routes).
- **Adversarial probes PASS** (evidence in `.omo/evidence/`): malformed input, stale-state, misleading-output, prompt-injection, hung-command timeout.

## Release Impact

Change type (tick one):

- [ ] Patch (fix)
- [X] Minor (feature)
- [ ] Major (breaking)

## Checklist

- [ ] **Changeset added** — I have run `pnpm changeset` (or `pnpm changeset --empty` if no CHANGELOG line is warranted) and committed the resulting `.md` file. Required for PRs targeting `main` and `release/*`; enforced by the `Changeset check` workflow.

## Reviewer Setup

**MSW (Mock Service Worker)**

- [ ] MSW must be enabled (`VITE_MSW_ENABLED=true` in `.env.local`)
- [ ] MSW not required
- [ ] Not Applicable

**Backend**

- [ ] No specific backend branch required
- [ ] Not Applicable

_Describe where in the UI the changes are visible and how to navigate there (if applicable). Paste a direct URL to the new page/component if applicable (e.g. `http://localhost:5173/...`):_

> **Example:** To test restarting an instance — navigate to the home page → click **Instances** → select an instance → open the **Operations** action menu → click **Restart**.
>
> _describe here_

**Testing Instructions** _(optional)_

_Describe any special testing instructions:_

1.

**Screenshots** _(optional)_

---

## Versioning Reminder

> [!IMPORTANT]
>


src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+    .map((entry) => entry.toString())
+    .filter((entry) => entry.endsWith(".spec.ts"))
+    .map((entry) => path.join(specDir, entry))
+    .sort();
+}
+
+function templateToPattern(expression: ts.TemplateExpression): string {
+  let pattern = expression.head.text;
+  for (const span of expression.templateSpans) {
+    pattern += `{param}${span.literal.text}`;
```
yurii-vasyliev: `templateToPattern` flattens every `${...}` span to `{param}`, and `patternToRegExp`
compiles `{param}` to `[^/:]+` — one segment, can't cross a `/`.

`local-repositories.spec.ts:67` does ``request.delete(`/v1/${createdRepoName}`)`` where
`createdRepoName` is `locals/<id>` (the spec asserts `body.name` contains `"locals/"` a few
lines up). We extract `/v1/{param}`; the declared route is `/v1/locals/{local}`. The
placeholder hides a segment boundary that exists at runtime, so it matches nothing and the
route reports as an uncovered gap even though this spec covers it. This one survives the
openapi bump — it's the only remaining orphan on `0.0.83`.

One caution: I tried the obvious fix — emitting `{param...}`, or trying both forms and
counting either as a match — and it's worse. Letting `{param}` span `/` makes
`DELETE /v1/{param}` match **4** declared routes and `GET /v1/{param}` match **11**. You'd
convert one visible false gap into several silent false *covered* results, and a gap that
never gets reported is invisible.

From the call site alone this URL is genuinely ambiguous, because the discriminating segment
(`locals`) lives inside the variable. So I'd suggest:

1. Land B3's orphan warning, which makes this loud instead of silent, and
2. Make the spec statically matchable — have it build the path from a literal prefix, e.g.
   ``request.delete(`/v1/locals/${shortName}`)`` — or add an explicit route-override
   annotation for the cases where that isn't possible.

Happy with any of those; the important thing is not to paper over it in the matcher.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+  calls: ExtractedCall[],
+): GapEntry[] {
+  const gaps: GapEntry[] = [];
+  for (const [routeId, metric] of Object.entries(report.routes)) {
+    const covered = calls.some(
+      (call) =>
+        call.method === metric.method.toUpperCase() &&
+        matchesPattern(call.urlPattern, metric.pattern),
+    );
+    if (!covered) {
```
yurii-vasyliev: When an extracted call matches no declared route it's just dropped, so the report can't
distinguish "the frontend covers this" from "our matcher is broken". Running extraction on
`main` today returns `warnings: []` while there are 3 genuine orphans.

This is the highest-value item in the review: it's the check that would have surfaced both
B1 and B2 without anyone reading the matcher, and it's what makes it safe to leave B2 as a
spec-side fix. An orphan is always a bug in the pin, the matcher, or the spec — never a
normal condition — so I'd surface it in `gaps.json` and fail the step (or at minimum annotate
the run) when the count is non-zero.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+}
+
+function listSpecFiles(specDir: string): string[] {
+  if (!fs.existsSync(specDir)) {
+    return [];
+  }
+  return fs
+    .readdirSync(specDir, { recursive: true })
+    .map((entry) => entry.toString())
+    .filter((entry) => entry.endsWith(".spec.ts"))
```
yurii-vasyliev: Only `*.spec.ts` is scanned, so any request issued from a shared helper is invisible to the
extractor. `POST /api/v2/login` is exercised by effectively every spec through
`helpers/auth.ts:12`, and it's absent from the extraction (7 calls / 5 distinct routes total).

```suggestion
    .filter((entry) => entry.endsWith(".ts") && !entry.endsWith(".d.ts"))
```

Note this alone isn't enough — the scan root also has to move up a level; see my other
comment on line 234.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+  specDir: string;
+  out: string;
+}
+
+function parseArgs(argv: string[]): CliOptions {
+  const options: CliOptions = {
+    report: REPORT_PATH,
+    specDir: path.resolve(
+      import.meta.dirname,
+      "../../../../e2e/docker-stack/api",
```
yurii-vasyliev: Second half of B4: `specDir` defaults to `e2e/docker-stack/api`, but `helpers/` is a
*sibling* of `api/`, not a child — so widening the extension filter on line 77 changes
nothing on its own. `readdirSync` is already recursive, so moving the root up one level is
all that's needed.

```suggestion
      "../../../../e2e/docker-stack",
```

If that pulls in more than you want, an explicit `["api", "helpers"]` root list would work
just as well.

src/tests/contract-coverage/eval-loop/prompt.ts
```diff
@@ -0,0 +1,57 @@
...
+  rank: number;
+}
+
+export interface SuggestionPrompt {
+  system: string;
+  user: string;
+}
+
+const MAX_GAPS = 5;
+const MAX_USER_CHARS = 12_000;
```
yurii-vasyliev: This ceiling is reachable on ordinary input, and it throws rather than degrading.

`computeGaps` copies `metric.contracts` straight into each gap (`collect-gaps.ts:183`), and
those contracts carry **full, untruncated** request *and* response payloads
(`aggregate.ts:142-148`). Ranking then puts mutating methods first (`collect-gaps.ts:191-193`)
— exactly the routes that carry contracts — so the top-5 slice is systematically the
heaviest entries.

Measured with one modest Landscape object (~430 bytes) and three statuses per route:

| gaps | user prompt | result |
| :--- | :--- | :--- |
| 1 | 7,546 chars | ok |
| 2 | 12,212 chars | **throws** |
| 5 | 26,210 chars | **throws** |

`MAX_GAPS` is 5, so in practice this raises every time there's more than one mutating gap.
I can see from `prompt.test.ts:90` that the guard is deliberate, and I agree with having one
— the issue is only that it trips on normal input and that `run-eval-loop.ts:90` doesn't
catch it, so the job ends on an unhandled stack trace.

Credit where due: `if: always()` on the artifact upload means `gaps.json` still survives, so
the design promise that LLM failures never lose deterministic data does hold. But the LLM
half of the loop would never actually run.

Suggest truncating payloads before they reach the prompt (one exemplar per status, or
hashes plus a short excerpt) and degrading — drop the lowest-ranked gap until it fits —
rather than throwing.

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
```diff
@@ -0,0 +1,189 @@
...
+    };
+  }
+
+  const client = options.client ?? createLlmClientFromEnv(options.env);
+  const exemplar = fs
+    .readFileSync(options.exemplarPath, "utf-8")
+    .split("\n")
+    .slice(0, EXEMPLAR_MAX_LINES)
+    .join("\n");
+  const prompt = buildSuggestionPrompt(gaps, exemplar);
```
yurii-vasyliev: Companion to my note on `prompt.ts:24`: `buildSuggestionPrompt` can throw ("prompt too
large") and there's no `try` around it, so the failure surfaces as an unhandled rejection
rather than the tidy `llm-failure` status with the raw-fallback file that you already handle
just below for unparseable responses. Worth routing it through the same path.

src/tests/contract-coverage/eval-loop/llm-client.ts
```diff
@@ -0,0 +1,163 @@
...
+        },
+        body: JSON.stringify({
+          model,
+          messages: [
+            { role: "system", content: req.system },
+            { role: "user", content: req.user },
+          ],
+          temperature: 0,
+          max_tokens: req.maxTokens ?? DEFAULT_MAX_TOKENS,
+        }),
```
yurii-vasyliev: No `response_format`, so strict JSON is requested in the system prompt but never enforced by
the API.

To be fair to the current code, this is less severe than it looks: `parseSuggestions`
(`render.ts:24`) already strips code fences, so a fenced reply survives. It's specifically
prose around the JSON — "Sure, here are five suggestions:" — that drops the whole batch.
Cheap to close properly:

```suggestion
        body: JSON.stringify({
          model,
          messages: [
            { role: "system", content: req.system },
            { role: "user", content: req.user },
          ],
          temperature: 0,
          response_format: { type: "json_object" },
          max_tokens: req.maxTokens ?? DEFAULT_MAX_TOKENS,
        }),
```

src/tests/contract-coverage/eval-loop/llm-client.ts
```diff
@@ -0,0 +1,163 @@
...
+  baseUrl: string;
+  model: string;
+  fetchImpl?: typeof fetch;
+  timeoutMs?: number;
+  backoffMs?: number;
+}
+
+const DEFAULT_TIMEOUT_MS = 60_000;
+const DEFAULT_BACKOFF_MS = 2_000;
+const DEFAULT_MAX_TOKENS = 4096;
```
yurii-vasyliev: 4096 output tokens is tight for the stated job: up to 5 suggestions, each a *complete
runnable Playwright spec file*. The exemplar alone is ~80 lines. Realistically this truncates
mid-JSON, which then fails `parseSuggestions` and discards the entire response — so the
failure mode is all-or-nothing rather than "fewer suggestions".

```suggestion
const DEFAULT_MAX_TOKENS = 16384;
```

src/tests/contract-coverage/eval-loop/types.ts
```diff
@@ -0,0 +1,221 @@
...
+  const response = expectRecord(value, "suggestionsResponse");
+  const suggestions = expectArray(
+    response.suggestions,
+    "suggestionsResponse.suggestions",
+  );
+  if (suggestions.length > 5) {
+    throw new Error(
+      "suggestionsResponse.suggestions: expected at most 5 items",
+    );
+  }
```
yurii-vasyliev: If the model returns 6 suggestions, this throws, `parseSuggestions` returns `null`, and
**all six are discarded** — a strictly worse outcome than keeping the first five. The cap is
already enforced where it matters: `render.ts:111` does `.slice(0, MAX_SUGGESTIONS)`. So
this check is redundant as well as destructive, and I'd just drop it:

```suggestion
```

(If you'd rather keep an explicit clamp here, `suggestions.length = 5;` in place of the
throw does the same job non-destructively.)

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+      fs.readFileSync(file, "utf-8"),
+      ts.ScriptTarget.Latest,
+      true,
+    );
+
+    const locationOf = (node: ts.Node): string => {
+      const { line } = sourceFile.getLineAndCharacterOfPosition(
+        node.getStart(sourceFile),
+      );
+      return `${path.basename(file)}:${line + 1}`;
```
yurii-vasyliev: `path.basename(file)` (also lines 137 and 144) drops the directory, so same-named specs in
different subdirectories become indistinguishable in the report — and this gets more likely
once B4 widens the scan. `path.relative(specDir, file)` keeps refs unique and clickable.
Copilot raised this one and the comment was suppressed.

src/tests/contract-coverage/eval-loop/llm-client.ts
```diff
@@ -0,0 +1,163 @@
+/**
+ * Zero-dependency LLM client for the eval loop. One OpenAI-compatible client
+ * shape covers GitHub Models today and OpenRouter/Gemini later — switching
```
yurii-vasyliev: Stale header: "covers GitHub Models today and OpenRouter/Gemini later", but
`DEFAULT_BASE_URL` (line 137) is OpenRouter and the workflow header explains why. Also a
suppressed Copilot finding.

.github/workflows/api-contract-eval.yml
```diff
@@ -0,0 +1,105 @@
...
+          NODE_OPTIONS: --max_old_space_size=4096
+          VITE_API_URL: ${{ vars.API_URL }}
+          VITE_API_URL_OLD: ${{ vars.API_URL_OLD }}
+          VITE_API_URL_DEB_ARCHIVE: ${{ vars.API_URL_DEB_ARCHIVE }}
+          VITE_API_PROXY_TARGET: ${{ vars.API_PROXY_TARGET }}
+          VITE_APP_TITLE: Landscape
+          VITE_ROOT_PATH: /new_dashboard/
+          VITE_APP_VERSION: "26.04.0.12"
+          VITE_APP_COMMIT: ${{ github.sha }}
+          VITE_SELF_HOSTED_ENV: true
```
yurii-vasyliev: These 9 variables are already set at job level (line 43) and job-level `env` is inherited,
so the re-declaration can go.

Worth keeping `NODE_OPTIONS` on line 76 though — that one is *not* set at job level, and it's
the 4 GB heap bump the full coverage run needs, so anchor the deletion at 77 rather than
removing the whole block.

```suggestion
```

Separately: the job block adds `VITE_REPORT_VIEW_ENABLED` and
`VITE_DETAILED_UPGRADES_VIEW_ENABLED`, which `tics-full.yml` doesn't have — so "env mirrors
the unit-test workflow" is off by two.

docs/testing/coverage.md
```diff
@@ -38,6 +38,19 @@ The repo already defines:
...
+  - **Evaluation (LLM):** `llm-client.ts` (zero-dependency, OpenAI-compatible fetch client; `LLM_API_KEY`/`LLM_BASE_URL`/`LLM_MODEL` env-driven, defaults to OpenRouter — GitHub Models entered retirement brownouts in July 2026) and `prompt.ts` (minimal prompt: top-5 gaps + their observed payloads as inert data + one exemplar spec, strict-JSON output contract, 12k-char guard) draft up to 5 Playwright API-contract test suggestions.
+
+Design rules for this area: gap detection must stay deterministic (scripts, not LLM); the LLM only elaborates already-computed gaps; no new npm dependencies; providers must be swappable via environment variables alone.
+
+Status: collection, LLM client, prompt builder, renderer, and the orchestrator are implemented and unit-tested (`eval:collect` / `eval:suggest` / `eval:full` pnpm scripts are wired). The scheduled GitHub Actions workflow is the remaining piece (see the eval-loop work plan).
```
yurii-vasyliev: Says the scheduled workflow "is the remaining piece", but it ships in this PR and the PR body
says nothing is deferred. Worth updating in the same pass so the doc doesn't immediately
contradict the code.

.prettierignore
```diff
@@ -1,4 +1,5 @@
 node_modules
 dist
 dist-ssr
-src/**/*.module.scss.d.ts
\ No newline at end of file
+src/**/*.module.scss.d.ts
+src/tests/contract-coverage/eval-loop/fixtures/malformed-report.json
```
yurii-vasyliev: Still no trailing newline — the change just moves the problem down a line.

yurii-vasyliev (changes requested):
Really solid piece of work, and the architecture is right. The Collection/Evaluation split
the design doc asks for is implemented properly: gap detection is fully deterministic and
the LLM only elaborates gaps the scripts already computed. Zero new dependencies, 49 unit
tests green (I ran them), all 8 CI checks passing, `gaps.json` always written before any LLM
call so an LLM failure never loses the deterministic data, `permissions: contents: read`,
and the actions are SHA-pinned — better practice than `tics-full.yml`, which pins by tag.
The developer README is genuinely good.

Requesting changes on the matching layer. Net effect today is that **the Go/debarchive layer
reports 100% false gaps**, including two endpoints `local-repositories.spec.ts` demonstrably
covers. One of the causes is a stale dependency pin rather than anything you wrote.

**Blocking**

- **B1 — stale openapi pin.** `@canonical/landscape-openapi` is pinned to `0.0.80`, which
  declares none of the routes the Go backend actually serves. Matched against the PR's own
  `matchesPattern`: `0.0.80` covers **0 of 30** routes, `0.0.83` covers **2 of 30**. Process
  fix, not code — bump it separately, or ship manual-only until that lands.
- **B2 — path parameters that span a `/`.** `templateToPattern` flattens every `${...}` to
  `{param}`, which compiles to `[^/:]+` (single segment). `local-repositories.spec.ts:67`
  does ``request.delete(`/v1/${createdRepoName}`)`` where `createdRepoName` is `locals/<id>`,
  so the real request has a segment the placeholder hides and the call matches nothing.
  Survives the pin bump — it's the last remaining orphan on `0.0.83`. Please **don't** fix
  this by making `{param}` multi-segment; see my inline note, I tested it and it over-matches.
- **B3 — no warning on orphan calls.** `computeGaps` silently drops extracted calls that
  match no declared route. On the current pin that's 3 genuine orphans and `warnings: []`.
  This is the check that would have surfaced B1 and B2 on its own, and it's what makes B2
  safe to leave unfixed in code.
- **B4 — shared helpers are invisible.** `listSpecFiles` scans only `*.spec.ts` under
  `e2e/docker-stack/api`, so `POST /api/v2/login` — exercised by every spec via
  `helpers/auth.ts:12` — never appears (extraction yields 7 calls / 5 routes). Needs both a
  wider filter and a wider scan root; details inline.
- **B5 — the prompt builder throws on realistic input, uncaught.** `MAX_USER_CHARS = 12_000`
  vs. gaps that carry full untruncated request/response payloads. Measured with one modest
  Landscape object and three statuses per route: 1 gap = 7,546 chars, **2 gaps = 12,212 →
  throws**, and the design targets 5. `run-eval-loop.ts:90` doesn't catch it. `if: always()`
  saves `gaps.json` so nothing is lost, but the LLM half of the loop never runs on real data.

**Cheap, same pass:** L1 `response_format`, L2 `max_tokens`, L3 clamp-don't-reject, and the
nits — all inline with suggestions.

**Not blocking, and my call rather than yours:** weekly cron vs. the doc's manual trigger,
the missing `## Copilot Prompt` section, and the `Caller` field. I'm fine with all three of
your choices — the design doc is what's out of date, and its "Vitest + Supertest" line is
simply wrong, so you were right to use Playwright. I'll update the doc.

All of the above is reproducible without running the full suite: `createOpenApiSource` takes
an injected spec, so both pin versions can be listed and matched with your own
`matchesPattern`.

package.json
yurii-vasyliev: This pin is the single biggest cause of the false-gap result. Matched with this PR's own
`matchesPattern`, `0.0.80` covers **0 of 30** debarchive routes; `0.0.83` covers **2 of 30**.
`0.0.80` simply predates the routes the Go service serves today.

Worth bumping in its own PR so this one isn't blocked on it — but until it lands, every
debarchive gap this loop reports is noise, so I'd keep the workflow manual-only rather than
let a weekly cron file suggestions against it.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+    );
+    if (!covered) {
+      gaps.push({
+        routeId,
+        method: metric.method,
+        pattern: metric.pattern,
+        backend: metric.backend,
+        totalHits: metric.totalHits,
+        statuses: metric.statuses,
+        ...(metric.contracts ? { contracts: metric.contracts } : {}),
```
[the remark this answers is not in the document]
marqode: added redaction helper and test file

package.json
marqode: Good catch, I opened a separate PR against main setting it to `0.0.83` in #789

src/tests/contract-coverage/sources/openapi.ts
```diff
@@ -71,7 +71,7 @@ function resolveWildcardParam(
  * Connector for the Go backend's real contract: the OpenAPI document shipped
  * in @canonical/landscape-openapi. `basePath` is the prefix the service is
  * mounted under in the frontend (VITE_API_URL_DEB_ARCHIVE minus the spec's
- * own /v1beta1 prefix).
+ * own /v1 prefix).
```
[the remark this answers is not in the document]
marqode: Rejected, better to swap out all /v1beta1 uses ahead of the openapi version bump in #789 which should land ahead of this

src/tests/contract-coverage/eval-loop/prompt.ts
marqode: Updated to truncate request / response payloads and summarize for large responses. If still too, large drops the lowest ranks gap.

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
marqode: wrapped in try/ catch with an informative error

src/tests/contract-coverage/eval-loop/types.ts
```diff
@@ -0,0 +1,216 @@
...
+}
+
+export function assertSuggestionsResponse(
+  value: unknown,
+): asserts value is SuggestionsResponse {
+  const response = expectRecord(value, "suggestionsResponse");
+  const suggestions = expectArray(
+    response.suggestions,
+    "suggestionsResponse.suggestions",
+  );
```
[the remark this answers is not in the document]
marqode: dropped based on @yurii-vasyliev's suggestion, already enforced elsewhere

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
```diff
@@ -0,0 +1,199 @@
...
+  return options;
+}
+
+async function main(): Promise<void> {
+  try {
+    const options = parseArgs(process.argv.slice(2));
+    const client =
+      process.env.LLM_MOCK === "1"
+        ? createMockClient(MOCK_LLM_RESPONSE)
+        : undefined;
```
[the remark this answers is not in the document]
marqode: use `RunOptions.mockFromGaps` to build `gaps.slice(0,5)`

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,280 @@
...
+  out: string;
+}
+
+function parseArgs(argv: string[]): CliOptions {
+  const options: CliOptions = {
+    report: REPORT_PATH,
+    specDir: path.resolve(
+      import.meta.dirname,
+      "../../../../e2e/docker-stack",
+    ),
```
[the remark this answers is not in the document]
marqode: changed based on previous PR comment to include helpers

src/tests/contract-coverage/eval-loop/collect-gaps.ts
marqode: addressed in [this commit](https://github.com/canonical/landscape-ui/pull/728/changes/8f246eb80f1285b811cc5d65209fa68dddd1f49a)

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,331 @@
...
+}
+
+const HTTP_METHODS = new Set([
+  "get",
+  "post",
+  "put",
+  "patch",
+  "delete",
+  "head",
+  "fetch",
```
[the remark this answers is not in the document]
marqode: removed `FETCH`, decide whether we want to derive the method from the second argument in a follow-up PR

src/tests/contract-coverage/eval-loop/collect-gaps.ts
marqode: Orphan labelling has been implemented. Handling multi-segment paths correctly and / or revising existing test specs to use literal path prefixes is a larger change that should land in a future PR to prevent further scope creep.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,275 @@
...
+    .filter((entry) => entry.endsWith(".spec.ts"))
+    .map((entry) => path.join(specDir, entry))
+    .sort();
+}
+
+function templateToPattern(expression: ts.TemplateExpression): string {
+  let pattern = expression.head.text;
+  for (const span of expression.templateSpans) {
+    pattern += `{param}${span.literal.text}`;
+  }
```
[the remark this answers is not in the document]
marqode: see above, to be addressed in follow-up PR

src/tests/contract-coverage/eval-loop/collect-gaps.ts
marqode: updated all 3 to use `path.relative()`

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  const parsed = parseSuggestions(completion.text);
+  const eligibleRoutes = new Set(
+    gaps.slice(0, 5).map(({ routeId }) => routeId),
+  );
+  if (
+    !parsed ||
+    parsed.suggestions.some(({ route }) => !eligibleRoutes.has(route)) ||
+    new Set(parsed.suggestions.map(({ route }) => route)).size !==
+      parsed.suggestions.length
```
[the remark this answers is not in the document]
marqode: added route validation

src/tests/contract-coverage/redact.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Recursively redact values whose keys contain sensitive tokens.
+ *
+ * Used before persisting MSW interactions or transmitting contract payloads
+ * to an external LLM, so credentials/tokens/passwords cannot leak into
+ * artifacts or leave the CI environment.
+ */
+export function redactSensitiveFields(payload: unknown): unknown {
+  if (payload === null || typeof payload !== "object") {
+    return payload;
```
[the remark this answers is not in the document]
marqode: added object conversion to extractPayload() in setup.ts

src/tests/contract-coverage/redact.ts
```diff
@@ -0,0 +1,36 @@
+const SENSITIVE_KEY_PATTERN =
+  /password|secret|token|apikey|api_key|authorization|privatekey|private_key|credential|accesskey|access_key|secretkey|secret_key|fde_recovery_key|recoverykey|signature|invitation_?id|secure_?id/i;
```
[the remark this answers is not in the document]
marqode: removed `signature`

marqode:
@yurii-vasyliev thanks for the thorough review. All those should be addressed now, with one change scoped to a follow-up PR. Let me know if you'd like me to split this PR up to make it easier to review.

src/tests/contract-coverage/eval-loop/types.ts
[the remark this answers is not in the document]
marqode: run-eval-loop.ts:158-177 now calls areSameRoutes(parsedRoutes, prompt.includedRoutes) after parsing, and the existing test "rejects an empty suggestions array" (line 219-226) passes today — an empty array fails the route-set-length comparison and returns llm-failure. Marking as closed

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,340 @@
...
+      console.error(
+        `[-] ${orphans.length} extracted spec call(s) match no declared route (orphans). ` +
+          "This indicates a matcher bug, stale route pin, or spec error.",
+      );
+      for (const orphan of orphans) {
+        console.error(
+          `    ${orphan.method} ${orphan.urlPattern} (${orphan.file}:${orphan.line})`,
+        );
+      }
+      process.exit(1);
```
RyanLoi98: Hey: I was looking at this section of code and your comments with Yurii about it and I wanted to run this by Copilot just to check the fixes implemented and if there were still any remaining bugs and this is what it said:
------------------------------------


**Blocking — the scheduled workflow will fail on every run.**

No objection to deferring the multi-segment fix (B2) — that's a reasonable scope call. The
issue is that B3 shipped as a hard `process.exit(1)` in the same PR, and those two decisions
don't compose.

The B3 comment asked to "surface it in `gaps.json` and fail the step **(or at minimum
annotate the run)**", and noted the orphan warning "is what makes it safe to leave B2 as a
spec-side fix" — i.e. a *warning* is what makes the deferral safe. A hard gate only becomes
safe once B2 lands.

I ran this PR's own extractor against the real `e2e/docker-stack` tree with a best-case
report, where every declared route is marked exercised. Three orphans remain:

```
DELETE /v1/{param}  (api/local-repositories.spec.ts:67)
DELETE /v1/{param}  (ui/mirrors/mirrors-crud.integration.spec.ts:84)
DELETE /v1/{param}  (ui/publication-targets/publication-targets-crud.integration.spec.ts:78)
```

Since this line exits 1 whenever `orphans.length > 0`, `pnpm eval:collect` fails, the
"Collect gaps (deterministic)" step fails, and `eval:suggest` never runs — every Monday.
The README already documents orphans as an expected limitation, which makes the fatal gate
inconsistent with the documented behaviour.

Worth noting this isn't resolved by #789: one orphan survives on `0.0.83` per the earlier
measurement in this review, and the gate trips on any count above zero.

Smallest fix that preserves your scope call: keep B2 deferred, and downgrade this to a
non-zero-count warning plus a step-summary annotation. The orphans still get surfaced loudly
in `gaps.json` and the run summary — which was the point of B3 — without blocking the job.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,340 @@
...
+}
+
+function listSpecFiles(specDir: string): string[] {
+  if (!fs.existsSync(specDir)) {
+    return [];
+  }
+  return fs
+    .readdirSync(specDir, { recursive: true })
+    .map((entry) => entry.toString())
+    .filter((entry) => entry.endsWith(".ts") && !entry.endsWith(".d.ts"))
```
RyanLoi98: I was also reading your discussion with Yurii on this (B4 I believe it was labeled) and Copilot brought up that solving that problem might have brought in a UI contaimination issue:


----

**Blocking — `ui/` integration specs are being counted as API-contract coverage.**

This was previously closed with "changed based on previous PR comment to include helpers".
Widening the scan root to all of `e2e/docker-stack` did solve the `helpers/auth.ts` problem,
but it introduced a new one: this recursive `.ts` filter also picks up `ui/**` integration
specs. 10 of the 21 extracted calls now come from `ui/`.

Those specs make direct `request.*` calls for setup and teardown, but they are not
API-contract tests. Counting them marks endpoints as covered when no contract test exists.

Measured by running `computeGaps` against an identical report with the two scan roots:

```
gaps when scanning whole docker-stack: 26
gaps when scanning api/ only        : 28

Hidden by counting ui/ specs as coverage:
  GET /debarchive/v1beta1/mirrors
  GET /debarchive/v1beta1/publicationTargets
```

Neither route has an API-contract test, so the loop reports them as covered. That's a silent
false negative in the tool's primary output — the failure mode flagged earlier in this
review as the worst kind, since a gap that never gets reported is invisible.

It also contradicts the docs shipped in this PR: both the eval-loop README diagram and
`docs/testing/coverage.md` state the scan covers `e2e/docker-stack/api/**/*.spec.ts`.

An explicit root list — `["api", "helpers"]` — keeps the `auth.ts` fix without pulling in `ui/`.

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
```diff
@@ -0,0 +1,264 @@
...
+  orphansFound: number;
+  suggestionsWritten: string[];
+  rawFallbackPath?: string;
+}
+
+const EXEMPLAR_MAX_LINES = 200;
+const RAW_FALLBACK_NAME = "00-raw-llm-output.md";
+
+function buildMockResponseFromGaps(gaps: GapEntry[]): string {
+  const suggestions = gaps.slice(0, 5).map(({ routeId }) => ({
```
RyanLoi98: I think this is the last one, but I was again reading the discussion between you and Yurii (really good points were brought up BTW so nice!) and while looking at those with copilot, I think a new bug was introduced when fixing the other 2.


----


**Non-blocking — `LLM_MOCK=1` fails whenever the prompt degrades.**

`buildMockResponseFromGaps` always emits 5 routes from `gaps`, but the response is validated
against `prompt.includedRoutes` (line 158), which is variable-length after degradation. The
mock client is built on line 130, before the prompt exists on line 136, so it can't know how
many routes actually made it in.

`areSameRoutes` starts with a length check, so as soon as `buildSuggestionPrompt` drops a gap
to fit the budget, the documented dry-run returns `llm-failure` and exits 1 instead of
exercising the renderer. Reachable on ordinary input — 3 statuses with ~400-char payloads:

```
statuses=3 payload=400  includedRoutes=3  ->  llm-failure
statuses=6 payload=400  includedRoutes=1  ->  llm-failure
```

This is an interaction between three fixes that are each correct on their own: mock-from-gaps
(always 5), validate-against-prompt, and degrade-instead-of-throw (which made the prompt
variable-length). The first two only agree while the third does nothing.

Fix: build the prompt first, then derive the mock from `prompt.includedRoutes` rather than
from `gaps`.

Scope is limited to the documented no-key dry-run — CI uses a real key, and `gaps.json` is
unaffected either way.

RyanLoi98 (changes requested):
Great work Marc, this PR is really really interesting and I think an excellent idea!


I don't have any major issues to bring up, and I don't have anything to add in terms of ideas or features.



On a finer level when looking at the stuff you and Yurii were talking about and trying to analyze things a bit deeper with copilot I think there are possible bugs that you should look at. I commented on them so you can look at it.



Other than that great work!

src/tests/contract-coverage/eval-loop/collect-gaps.ts
marqode: Good catch, I'll downgrade the orphan exit to a warning so it doesn't stop the run

src/tests/contract-coverage/eval-loop/collect-gaps.ts
marqode: Valid, I'll add the explicit root list

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
marqode: Moved mock-client construction to after buildSuggestionPrompt runs, and changed the mock builder from buildMockResponseFromGaps(gaps) (always 5 routes) to buildMockResponseFromRoutes(prompt.includedRoutes), so the mock response always matches whatever routes actually survived prompt-size degradation. Added a regression test (LLM_MOCK=1 succeeds even when the prompt degrades the gap list) using an 11k-char exemplar that forces route-dropping, asserting status === "ok"

.github/copilot-instructions.md
```diff
@@ -59,6 +59,8 @@ pnpm dev       # Vite dev server (5173)
 pnpm build     # Lint + typecheck + Vite build
 pnpm vitest    # Unit/integration tests
 pnpm test      # Playwright E2E
+pnpm eval:collect  # API-contract gap report (needs `pnpm coverage:full` first)
+pnpm eval:suggest  # LLM-drafted spec suggestions (needs LLM_API_KEY, or LLM_MOCK=1 for a dry-run)
```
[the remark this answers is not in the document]
marqode: Scope to future work

marqode:
@yurii-vasyliev @RyanLoi98 thanks for the reviews! All comments should be addressed at this point, and I updated the PR description and design doc to add some context for the limitations of this feature / potential next steps.

RyanLoi98:
> @yurii-vasyliev @RyanLoi98 thanks for the reviews! All comments should be addressed at this point, and I updated the PR description and design doc to add some context for the limitations of this feature / potential next steps.

Great work Marc! Im reviewing your PR now, but can you also fix the prettier error? That way when I and Yurii aprove it, you can merge it right away without having to fix the errors and then wait for another reapproval.

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
```diff
@@ -0,0 +1,270 @@
...
+  writeGapsFile(
+    buildGapsFile(report, extraction, gaps, orphans),
+    path.join(options.outDir, "gaps.json"),
+  );
+  fs.rmSync(path.join(options.outDir, "suggestions"), {
+    recursive: true,
+    force: true,
+  });
+
+  if (orphans.length > 0) {
```
yurii-vasyliev: blocking: this is the other half of the orphan gate. The collector now warns, but this branch still returns `orphans` and `main()` exits 1, so with the one `DELETE /v1/{param}` orphan on the current tree the Suggest step fails before any LLM call. Log the orphans here and carry on to the prompt, and flip the "stops before calling the client" test at `run-eval-loop.test.ts:118` to match.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
```diff
@@ -0,0 +1,357 @@
...
+    if (aMutating !== bMutating) return aMutating - bMutating;
+    return a.routeId.localeCompare(b.routeId);
+  });
+  return gaps.map((gap, index) => ({ ...gap, rank: index + 1 }));
+}
+
+/**
+ * Orphan = extracted spec call that matches no exercised route. Unlike a gap,
+ * an orphan means the matcher, the route pin, or the spec is wrong, so it is
+ * always treated as an error.
```
yurii-vasyliev: nit: stale now that orphans are a warning.

src/tests/contract-coverage/eval-loop/prompt.ts
```diff
@@ -0,0 +1,147 @@
...
+
+function truncatePayload(payload: unknown): unknown {
+  if (payload === null || payload === undefined) {
+    return payload;
+  }
+  const serialized = JSON.stringify(payload);
+  if (serialized.length <= MAX_PAYLOAD_CHARS) {
+    return payload;
+  }
+  return { __truncated: true, originalChars: serialized.length };
```
yurii-vasyliev: nit: swapping the payload for a marker loses the shape, which is what the model needs to write assertions. A 500 char prefix plus `originalChars` would keep it useful.

yurii-vasyliev (changes requested):
Thanks, nearly everything from the last round is in and the redaction work is a good addition. One thing left.

Blocking
- The orphan gate was only downgraded in `collect-gaps.ts`. `run-eval-loop.ts` still returns `orphans` and exits 1 before the LLM is called, and that is the script the "Suggest tests" step runs. The current spec tree has exactly one orphan, the `DELETE /v1/${createdRepoName}` call, and it survives the 0.0.83 bump, so the scheduled run fails every Monday and never produces a suggestion. Details inline.

src/tests/contract-coverage/eval-loop/run-eval-loop.ts
marqode: Fixed — run() now logs the orphan(s) and falls through to gap evaluation, so the run reaches the LLM even with the `DELETE /v1/${createdRepoName}` orphan. Removed "orphans" from RunResult["status"] and the dead process.exit(1) branch in main(), since orphans no longer produce a terminal status. Flipped run-eval-loop.test.ts:118 (now "orphans: logs a warning but still calls the client and writes suggestions") to exercise ORPHAN_SPEC_DIR and asserts the client is called and all 4 gap suggestions are written despite the 1 orphan.

src/tests/contract-coverage/eval-loop/prompt.ts
marqode: Changed truncatePayload() to keep a 500-char JSON prefix (preview) alongside originalChars instead of a bare marker, so the model still sees field names/types for oversized payloads. Covered by the existing prompt.test.ts truncation test.

src/tests/contract-coverage/eval-loop/collect-gaps.ts
marqode: Updated the docstring, now describes them as a warning reflecting a likely matcher limitation.

src/tests/contract-coverage/redact.ts
```diff
@@ -0,0 +1,39 @@
+// Excludes "has_"/"no_" (and camelCase "has"/"no") prefixes so boolean
+// presence flags like `has_password` or `hasToken` survive redaction,
+// while suffix matches like `AccessToken` are still caught.
+const SENSITIVE_KEY_PATTERN =
+  /(?<!has_)(?<!has)(?<!no_)(?<!no)(password|secret|token|apikey|api_key|authorization|privatekey|private_key|credential|accesskey|access_key|secretkey|secret_key|fde_recovery_key|recoverykey|invitation_?id|secure_?id|gpg_?key)/i;
```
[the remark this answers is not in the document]
marqode: now gated on the value actually being typeof === "boolean": sensitive keys are redacted unconditionally unless they're both boolean-flag-shaped and hold a real boolean. Added a regression test

src/tests/setup.ts
```diff
@@ -27,18 +29,31 @@ configure({ asyncUtilTimeout: 5000 });
...
     const clone = streamOwner.clone();
     const text = await clone.text();
-    try {
-      return JSON.parse(text);
-    } catch {
-      return text || null;
+    if (!text) return null;
+
+    const contentType = streamOwner.headers.get("content-type") ?? "";
+    if (contentType.includes("application/json")) {
```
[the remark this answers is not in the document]
marqode: Fixed — content-type matching now uses /\/(?:[\w.-]+\+)?json\b/i instead of a plain substring check, so application/problem+json, application/vnd.api+json, application/connect+json, etc. all get parsed to an object (so redactSensitiveFields can act on them) instead of falling through as an opaque raw string. Verified it doesn't false-positive on non-JSON types that merely contain "json" (e.g. a hypothetical application/jsonlines). Added 2 regression tests.

RyanLoi98 (dismissed):
LGTM!

